PR #36, opened 2015-01-20
Added middleware endpoint

AlfonsoUceda:
lotus/controller#72 Create a middleware endpoint for an action with one or several middlewares.

This is necessary for creating new PR in lotus/controller, fixing the rack.rb class.

Example:

```ruby
require 'lotus/router'
require 'lotus/controller'

class MyMiddleware
  def initialize(app)
   @app = app
  end

  def call(env)
    code, headers, body = @app.call(env)
    [code, headers.merge('X-Middleware' => 'CALLED'), body]
  end
end

module Home
  class Index
    include Lotus::Action
    use MyMiddleware

    def call(params)
      self.body = 'foo'
    end
  end
end

app = Lotus::Router.new do
  get '/', to: 'home#index'
end

Rack::Server.start app: app, Port: 2300
```
I haven't added new tests, ClassEndpoint isn't tested, how is it possible to test this MiddlewareEndpoint?

### DIFF
--- a/lib/lotus/routing/endpoint_resolver.rb
+++ b/lib/lotus/routing/endpoint_resolver.rb
@@ -183,11 +183,14 @@ module Lotus
       end
 
       def constantize(string)
-        begin
-          ClassEndpoint.new(Utils::Class.load!(string, @namespace))
-        rescue NameError
-          LazyEndpoint.new(string, @namespace)
+        klass = Utils::Class.load!(string, @namespace)
+        if klass.respond_to?(:call)
+          Endpoint.new(klass)
+        else
+          ClassEndpoint.new(klass)
         end
+      rescue NameError
+        LazyEndpoint.new(string, @namespace)
       end
 
       def classify(string)
--- a/test/integration/middleware_integration_test.rb
+++ b/test/integration/middleware_integration_test.rb
@@ -1,0 +1,26 @@
+require 'test_helper'
+
+describe 'Lotus middleware integration' do
+  before do
+    @routes = Lotus::Router.new(namespace: Web::Controllers) do
+      get '/',          to: 'home#index'
+      get '/dashboard', to: 'dashboard#index'
+    end
+
+    @app = Rack::MockRequest.new(@routes)
+  end
+
+  it 'action with middleware' do
+    response = @app.get('/')
+    response.body.must_equal 'Hello from Web::Controllers::Home::Index'
+    response.status.must_equal 200
+    response.headers.fetch('X-Middleware').must_equal 'CALLED'
+  end
+
+  it 'action without middleware' do
+    response = @app.get('/dashboard')
+    response.body.must_equal 'Hello from Web::Controllers::Dashboard::Index'
+    response.status.must_equal 200
+    response.headers['X-Middleware'].must_be_nil
+  end
+end
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -1,6 +1,32 @@
 require 'rexml/document'
 require 'lotus/routing/parsing/parser'
 
+module Web
+  module Controllers
+    module Home
+      class Index
+        # mocking class call method for middleware
+        def self.call(env)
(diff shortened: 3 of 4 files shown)


jodosha:
@AlfonsoUceda Thanks for this PR.

Please have a look at https://github.com/lotus/router/blob/master/test/routing/endpoint_resolver_test.rb

AlfonsoUceda:
@jodosha what do you think about these tests?

lib/lotus/routing/endpoint_resolver.rb
```diff
@@ -197,7 +198,7 @@ def classify(string)
       private
       def resolve_callable(callable)
         if callable.respond_to?(:call)
-          @endpoint_class.new(callable)
+          create_endpoint(callable, @endpoint_class)
```
AlfonsoUceda: @jodosha  do you think create_endpoint here is necessary? because always inside create_endpoint is checking again the respond_to and it always goes through `MiddlewareEndpoint`

jodosha: @AlfonsoUceda We can review it later, when the PR for Controller will be in place ;)

jodosha:
Things are simpler than I expected :wink:

There are two symptoms that led me to this finding:
  1. Code duplication: #create_endpoint and #resolve_callable are doing the same thing.
  2. We are testing implementation instead of behaviors: https://github.com/lotus/router/pull/36/files#diff-a3ea2e007d31d69e1fd367fc500263b2R65

I found that `MiddlewareEndpoint` isn't needed at all. `Endpoint` is enough.
Because it inheriths from `SimpleDelegator`, so all the missing methods are forwarded to the wrapped object.

```ruby
def call(env)
  __getobj__.call(env)
end
```

Can be replaced with

```ruby
def call(env)
  super
end
```

That means, we can remove it entirely.
We just be smart enough to pass the right object to `Endpoint`.

Now, the tests that we put in place aren't covering **what** we want to achieve, but **how** we want to achieve it.

In other words, what we want to do is expressed by high level tests, usually integration tests.
Where we simulate a Lotus action with an inner middleware, and a router that forwards HTTP requests to it.

That's basically our feature (the **"what"**).

Testing if a returned object is of a given type is useful in certain scenarios, but it's always focused about the details (the **"how"**).

To recap, we probably need another integration test here. :smile:

We still can use `Endpoint`
And probably we should **only** do this:

```ruby
def resolve_matchable(matchable)
  if matchable.respond_to?(:match)
    constantize(
      resolve_callable(resolve_action(matchable)) || classify(matchable)
    )
  end
end
```

I learned a great lesson. @AlfonsoUceda thanks for making this possible. :+1:

lib/lotus/routing/endpoint.rb
```diff
@@ -44,6 +44,11 @@ def inspect
           "#<#{ __getobj__.class }>"
         end
       end
+
+      # @since x.x.x
+      def call(env)
```
jodosha: This isn't necessary, because `Endpoint` inherits from `SimpleDelegator` this behavior **already happens**. ;)

test/integration/middleware_integration_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    response.body.must_equal 'Hello from Web::Controllers::Home::Index'
+    response.status.must_equal 200
+    response.headers.fetch('X-Middleware').must_equal 'CALLED'
+  end
+
+  it 'action without middleware' do
+    response = @app.get('/dashboard')
+    response.body.must_equal 'Hello from Web::Controllers::Dashboard::Index'
+    response.status.must_equal 200
+    -> { response.headers.fetch('X-Middleware') }.must_raise KeyError
```
jodosha: There is no need to raise an exception here, because we already know that the key is missing.
What if we change it with?

```ruby
response.headers['X-Middleware'].must_be_nil
```

:smile_cat:

jodosha:
This looks good :+1: Thank you!

PR #39, opened 2015-02-06
Removed support for MRI ruby versions lower than 2.2

AlfonsoUceda:
ditto


jodosha:
@AlfonsoUceda Thank you :+1:

PR #45, opened 2015-03-05
Controller name options for #resource/#resources

gzigzigzeo:
resource 'users', controller: 'profile'

Now produces:

get '/users' to: Web::Controller::Profile::Index
get '/users/new' to: Web::Controller::Profile::New

### DIFF
--- a/lib/lotus/routing/resource/action.rb
+++ b/lib/lotus/routing/resource/action.rb
@@ -206,7 +206,7 @@ module Lotus
         # @api private
         # @since 0.1.0
         def endpoint
-          [ resource_name, action_name ].join separator
+          [ controller_name, action_name ].join separator
         end
 
         # Separator between controller and action name
@@ -220,6 +220,22 @@ module Lotus
         # @since 0.1.0
         def separator
           @options[:separator]
+        end
+
+        # Resource controller name
+        #
+        # @example
+        #   Lotus::Router.new do
+        #     resources 'flowers', controller: 'rocks'
+        #   end
+        #
+        #   # It will mount path 'flowers/new' to Rocks::New instead of Flowers::New
+        #   # Same for other action names
+        #
+        # @api private
+        # @since 0.2.1
+        def controller_name
+          @options[:controller] || resource_name
         end
       end
 
@@ -249,7 +265,7 @@ module Lotus
         end
 
         def endpoint(action_name)
-          [ resource_name, action_name ].join separator
+          [ controller_name, action_name ].join separator
         end
 
         def as(action_name)
--- a/lib/lotus/routing/resource/options.rb
+++ b/lib/lotus/routing/resource/options.rb
@@ -21,6 +21,7 @@ module Lotus
         # @param options [Hash]
         # @option options [Hash] :only white list of the default actions
         # @option options [Hash] :except black list of the default actions
+        # @option options [String] :controller namespace for an actions
         #
         # @api private
         # @since 0.1.0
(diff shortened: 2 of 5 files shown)


jodosha:
@gzigzigzeo Thanks for this PR, can you please elaborate more your use case? Thank you.

gzigzigzeo:
Sure.

namespace 'admin' do
  resources :users, controller: 'admin/users'
end

gzigzigzeo:
@jodosha It is useful with namespacing support merged to utils.

jodosha:
@gzigzigzeo Sorry if I insist, but I'm trying to have a better understanding. By looking at the name of the action it looks like a Lotus convention, so I guess a greenfield project. Those kind of configurations are often required for legacy code.

So why do you have the need of decouple the name of the resource from the name of the controller?

gzigzigzeo:
@jodosha 

- Use case for the lastest example: 

```ruby
resource 'me', only: %i(show edit update), controller: 'profile'
```

- It should not break the integration because I can do the same thing manually, without use #resources:

```ruby
namespace 'admin' do
  resources 'people', controller: 'admin/users'
end
```

is equal to

```ruby
namespace 'admin' do
  get '/people', to: 'admin/users#index'
  get '/people/new', to: 'admin/users#new'
  ...
```

With my merged PR to `lotus/utils` (`#classify` fixes) it works fine.

Sure, convention is important and good. This option does not break the convention, but gives me more control on generated resource routes.

AlfonsoUceda:
if this doesn't break the convention and give us more flexibility then :+1:

joneslee85:
:+1: much thanks

lib/lotus/routing/resource/action.rb
```diff
@@ -221,6 +221,22 @@ def endpoint
...
+        # @example
+        #   Lotus::Router.new do
+        #     resources 'flowers', controller: 'rocks'
+        #   end
+        #
+        #   # It will mount path 'flowers/new' to Rocks::New instead of Flowers::New
+        #   # Same for other action names
+        #
+        # @api private
+        # @since 0.2.1
```
jodosha: @gzigzigzeo @joneslee85 this should be `@since x.x.x`

jodosha:
@gzigzigzeo Thank you very much for this. :+1:

jodosha:
@gzigzigzeo @joneslee85 This is a nice addition, but it works only for `resource`. We should add the same functionality for `resources` as well.

gzigzigzeo:
@jodosha

* Can not reproduce it with your snippet:

```ruby
Bundler could not find compatible versions for gem "lotus-utils":
  In Gemfile:
    lotus-controller (~> 0.3) ruby depends on
      lotus-utils (~> 0.4) ruby

    lotusrb (>= 0) ruby depends on
      lotus-utils (0.3.5)
```
I think lotus/lotus dependency on lotus-utils should be changed as follows:

```ruby
  spec.add_dependency 'lotus-utils',      '~> 0.4', '>= 0.4.0'
```

* You have the mistake in your example:

```ruby
resources :bo, controller: 'books', only: :show
```
creates single route path '/bo/:id' (it will work, I have checked it in my current app). To create '/bo' you should specify change the `:only` option as follows:

```ruby
resources :bo, controller: 'books', only: %i(show index)
```

lib/lotus/routing/resource/action.rb
gzigzigzeo: @jodosha Sorry, I did not understood. I see @since there, am I wrote wrong version number?

PR #57, opened 2015-06-04
Added force_ssl option

### LINKED ISSUE #56: Force SSL
Introduce a new option for `Lotus::Router` constructor: `force_ssl`. It accepts a boolean. When it's true and the router receives a non-encrypted request (`http`), it redirects to the secure equivalent resource (`https`). It should use a `301` HTTP status code.

AlfonsoUceda:
Resolves #56 

Now you can force ssl in your router:
```ruby
# notice scheme is http
@router = Lotus::Router.new(force_ssl: true) do
  get '/', to: 'home#index', as: :root
end
```
if you call to your app's home then `@router` return 301 status and Location `https://localhost/`

### DIFF
--- a/lib/lotus/routing/force_ssl.rb
+++ b/lib/lotus/routing/force_ssl.rb
@@ -1,0 +1,136 @@
+module Lotus
+  module Routing
+    # Force ssl
+    #
+    # Redirect response to the secure equivalent resource (https)
+    #
+    # @since x.x.x
+    # @api private
+    class ForceSsl
+      # Https scheme
+      #
+      # @since x.x.x
+      # @api private
+      SSL_SCHEME = 'https'.freeze
+
+      # Location header
+      #
+      # @since x.x.x
+      # @api private
+      LOCATION_HEADER = 'Location'.freeze
+
+      # Default http port
+      #
+      # @since x.x.x
+      # @api private
+      DEFAULT_HTTP_PORT = 80
+
+      # Default ssl port
+      #
+      # @since x.x.x
+      # @api private
+      DEFAULT_SSL_PORT = 443
+
+      # Moved Permanently http code
+      #
+      # @since x.x.x
+      # @api private
+      MOVED_PERMANENTLY_HTTP_CODE = 301
+
+      # Temporary Redirect http code
+      #
+      # @since x.x.x
+      # @api private
+      TEMPORARY_REDIRECT_HTTP_CODE = 307
+
+      # Initialize force ssl class.
+      #
+      # @param force_ssl [Boolean] activate redirection to ssl
+      # @param options [Hash] activate redirection to ssl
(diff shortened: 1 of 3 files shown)


lib/lotus/routing/force_ssl.rb
```diff
@@ -0,0 +1,93 @@
...
+      #
+      # @since x.x.x
+      # @api private
+      LOCATION_HEADER = 'Location'.freeze
+
+      # Default port number
+      #
+      # @since x.x.x
+      # @api private
+      DEFAULT_PORT = 80
```
jodosha: Default port for SSL is 443.

AlfonsoUceda: @jodosha I put 80 because I don't want a url like this:
`http://www.mydomain:80/path?query=string`

jodosha: @AlfonsoUceda When you use SSL via browser, it uses 443, even if you don't see it.

AlfonsoUceda:
@jodosha what do you think now?

lib/lotus/routing/http_router.rb
```diff
@@ -112,6 +115,18 @@ def raw_call(env, &blk)
...
 
+      # Resolve the given Rack env to a registered endpoint and invoke it.
+      #
+      # @param env [Hash] a Rack env instance
+      #
+      # @return [Rack::Response, Array]
+      #
+      # @since x.x.x
+      # @api private
+      def call(env)
```
jodosha: @AlfonsoUceda Can you please make this working in `#raw_call` instead of here? `http_router` does low level stuff with this method (undefining and redefining it), that's why `#raw_call` it's there.

jodosha:
@AlfonsoUceda One last thing before we merge this. `Lotus::Router` should honor `HTTP_X_FORWARDED_PROTO` header.

jodosha:
@AlfonsoUceda The current implementation always force SSL if `force_ssl: true` is given.

The problem is that when an incoming request that already uses SSL (scheme `https`), is redirected, which is wrong. In this case we should let the request in, otherwise we'll create an infinite redirect problem.

AlfonsoUceda:
@jodosha sorry I forgot add tests with force_ssl option set true and scheme https, you can check there isn't an infinite loop.

What do you want with `HTTP_X_FORWARDED_PROTO`? can you explain it? please

jodosha:
@AlfonsoUceda I've just discovered that `Rack::Request#scheme` already takes care of this. https://github.com/rack/rack/blob/e9bd27d5e3be10bddbb14064319c2cad24a036a2/lib/rack/request.rb#L72

AlfonsoUceda:
@jodosha what do you think now?

jodosha:
@AlfonsoUceda I think it's good. Merged :+1:

PR #58, opened 2015-06-08
Added prefix option to router.

AlfonsoUceda:
# Problem
There is a problem in lotusrb gem generating correct routes when the apps are mounted inside lotus container:
```ruby
class AdminApp  < Lotus::Application
  configure do
    routes do
      get '/home', to 'home#index', as: :home
    end
  end
end
Lotus::Container.configure do
  mount AdminApp, at: '/admin'
  mount App, at: '/'
end
AdminApp::Routes.path(:home) # => '/home' <- WRONG
# should be
AdminApp::Routes.path(:home) # => '/admin/home' <- GOOD
```

# Solution
This PR add prefix option to router, later we can pass in lotusrb the `at:` option as `prefix` for the router.

```ruby
router = Lotus::Router.new(prefix: '/admin') do
  get '/home', to: 'home#index', as: ':home'
end

router.path(:home) # => '/admin/home'
```

### DIFF
--- a/lib/lotus/router.rb
+++ b/lib/lotus/router.rb
@@ -341,7 +341,7 @@ module Lotus
     #    # It will map to Flowers::Index.new, which is the
     #    # Lotus::Controller convention.
     def get(path, options = {}, &blk)
-      @router.prefixed_namespace.get(path, options, &blk)
+      @router.get(path, options, &blk)
     end
 
     # Defines a route that accepts a POST request for the given path.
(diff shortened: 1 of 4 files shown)


lib/lotus/router.rb
```diff
@@ -341,7 +341,7 @@ def defined?
     #    # It will map to Flowers::Index.new, which is the
     #    # Lotus::Controller convention.
     def get(path, options = {}, &blk)
-      @router.get(path, options, &blk)
+      @router.prefixed_namespace.get(path, options, &blk)
```
jodosha: @AlfonsoUceda This leaks `Lotus::Routing::HttpRouter` details to the higher level. That low level class should hide this detail. `Lotus::Router#get` should remain untouched, while `Lotus::Routing::HttpRouter#get` should handle this.

jodosha:
@AlfonsoUceda This looks really, really clean now :clap:

PR #59, opened 2015-06-09
Fix docs for head route

triplepointfive:


jodosha:
@triplepointfive Thanks for this PR. :+1:

PR #61, opened 2015-06-25
Fix container routes bug

AlfonsoUceda:
Fix a bug with container arch.

There was a problem duplicating the `at` option in routes url.

### DIFF
--- a/lib/lotus/router.rb
+++ b/lib/lotus/router.rb
@@ -902,7 +902,7 @@ module Lotus
     #   router.path(:login, return_to: '/dashboard') # => "/login?return_to=%2Fdashboard"
     #   router.path(:framework, name: 'router')      # => "/router"
     def path(route, *args)
-      @router.path(route, *args)
+      @router.custom_path(route, *args)
     end
 
     # Generate a URL for a specified named route.
@@ -929,7 +929,7 @@ module Lotus
     #   router.url(:login, return_to: '/dashboard') # => "https://lotusrb.org/login?return_to=%2Fdashboard"
     #   router.url(:framework, name: 'router')      # => "https://lotusrb.org/router"
     def url(route, *args)
-      @router.url(route, *args)
+      @router.custom_url(route, *args)
     end
 
     # Returns an routes inspector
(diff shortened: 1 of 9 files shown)


test/prefix_test.rb
```diff
@@ -95,7 +95,7 @@ def call(env)
 
     %w(GET POST PUT PATCH DELETE TRACE OPTIONS).each do |verb|
       it "recognizes requests (#{ verb })" do
-        env = Rack::MockRequest.env_for('/admin/home', method: verb)
+        env = Rack::MockRequest.env_for('/home', method: verb)
```
jodosha: @AlfonsoUceda Why have you removed the prefix from the test?

jodosha:
@AlfonsoUceda This PR doesn't reproduce the problem, and it removes the prefix from the existing tests.

We don't have the proof that it solves the problem.

test/prefix_test.rb
AlfonsoUceda: ok, with the problem, now path_prefix is only for path and url generation, if we use like namespace we have problems with lotus architecture

Then we can use namespace if we want that namespace in the request url and path_prefix only for path and url helpers

AlfonsoUceda:
@jodosha what do you think to merge this? I'll create lotus PR testing this behaviour

lib/lotus/router.rb
```diff
@@ -902,7 +902,7 @@ def call(env)
     #   router.path(:login, return_to: '/dashboard') # => "/login?return_to=%2Fdashboard"
     #   router.path(:framework, name: 'router')      # => "/router"
     def path(route, *args)
-      @router.path(route, *args)
+      @router.custom_path(route, *args)
```
joneslee85: @AlfonsoUceda can you please explain why you renamed it to `custom_`?

joneslee85:
@AlfonsoUceda I've hand-tested this and it works nicely

lib/lotus/router.rb
AlfonsoUceda: it's very weird but `path` didn't call :S and the custom exception didn't raise, that's why I fix exception tests.

PR #64, opened 2015-06-30
PathPrefix relative_join bug

theocodes:
This is related to:

https://github.com/lotus/utils/pull/82

Please find there all the explanation.

Thanks!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,6 @@ if !ENV['TRAVIS']
   gem 'yard',   require: false
 end
 
-gem 'lotus-utils', '~> 0.5', require: false, github: 'lotus/utils', branch: '0.5.x'
+gem 'lotus-utils', '~> 0.5', require: false, github: 'lotus/utils', branch: 'master'
 gem 'simplecov',             require: false
 gem 'coveralls',             require: false
--- a/test/prefix_test.rb
+++ b/test/prefix_test.rb
@@ -40,6 +40,12 @@ describe Lotus::Router do
         trace   '/home', to: 'home#index', as: :trace_home
         options '/home', to: 'home#index', as: :options_home
 
+        get  '/admin',      to: 'home#index', as: :get_admin
+        get  '/admin/new',  to: 'home#index', as: :new_admin
+        get  '/admin/edit', to: 'home#index', as: :edit_admin
+        post '/admin',      to: 'home#index', as: :create_admin
+        put  '/admin',      to: 'home#index', as: :put_admin
+
         resources :users
         resource :asteroid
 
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -6,6 +6,6 @@ if !ENV['TRAVIS']
   gem 'yard',   require: false
 end
 
-gem 'lotus-utils', '~> 0.5', require: false, github: 'lotus/utils', branch: '0.5.x'
+gem 'lotus-utils', '~> 0.5', require: false, github: 'theocodes/utils', branch: 'same-strings-join-fix'
```
theocodes: I didn't change this back on purpose so people don't forget to point it at my fix on utils before running tests etc.

test/named_routes_test.rb
```diff
@@ -50,7 +50,7 @@
     end
 
     it 'recognizes glob string' do
-      @router.path(:glob).must_equal '/files/'
+      @router.path(:glob).must_equal '/files'
```
theocodes: This seemed like a insignificant change which i didn't it would effect anything else. But please if you know a reason why there should be that trailing slash there, let me know!

theocodes: @jodosha can you please confirm this?

AlfonsoUceda: that is because of this https://github.com/lotus/router/blob/master/lib/lotus/routing/http_router.rb#L48

maybe you can get rid that Utils::PathPrefix and modify this: 
https://github.com/lotus/router/blob/master/lib/lotus/routing/http_router.rb#L162
`options[:prefix] || ''`
using something like `@prefix.concat uri.path`

theocodes:
This has become irrelevant as the issue was fixed on utils.

AlfonsoUceda:
@theocodes but the tests added `/admin/admin` are correct because we wan't to get a regression with this.

theocodes:
@AlfonsoUceda yes, thanks! It's been amended to only add those tests now. This however, will obviously fail without the fix on utils :)

test/prefix_test.rb
```diff
@@ -40,6 +40,12 @@ def call(env)
         trace   '/home', to: 'home#index', as: :trace_home
         options '/home', to: 'home#index', as: :options_home
 
+        get     '/admin',      to: 'home#index',      as: :get_admin
```
AlfonsoUceda: could you remove unnecessary spaces?

test/prefix_test.rb
```diff
@@ -40,6 +40,12 @@ def call(env)
         trace   '/home', to: 'home#index', as: :trace_home
         options '/home', to: 'home#index', as: :options_home
 
+        get  '/admin',      to: 'home#index', as: :get_admin
```
theocodes: @AlfonsoUceda like that?

AlfonsoUceda: yes perfect!

Gemfile
AlfonsoUceda: @theocodes your lotus/utils PR was merged. Can you change this line with lotus/utils master branch? thanks ;)

jodosha:
@theocodes Wonderful. Merging this! :+1:

PR #70, opened 2015-10-16
Parse JSON Body parameters for requests made with JSON API Content-Type

stravid:
The [JSON API specification](http://jsonapi.org/format/#content-negotiation) requires all requests to use the `Content-Type` `application/vnd.api+json`. Since the parameters are sent as JSON in the body, Rack ignores them. Therefore the work has to be done by Lotus, this adds the content type to the `JsonParser` so it correctly parses the requests. Otherwise the mismatch of the content types lead to missing/ignored parameters.

### DIFF
--- a/lib/lotus/routing/parsing/json_parser.rb
+++ b/lib/lotus/routing/parsing/json_parser.rb
@@ -5,7 +5,7 @@ module Lotus
     module Parsing
       class JsonParser < Parser
         def mime_types
-          ['application/json']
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
to me :+1: this content type is registered in IANA: http://www.iana.org/assignments/media-types/application/vnd.api+json

jodosha:
@stravid Thank you :+1:

PR #74, opened 2015-11-16
Introduce Lotus::Router.define

jodosha:
Returns the given block as it is.

When `Lotus::Router` is used as a standalone gem and the routes are defined into a configuration file, some systems could raise an exception.

Imagine the following file into a Ruby on Rails application:

```ruby
# app/api/config/routes.rb
get '/', to: 'api#index'
```

Because Ruby on Rails in production mode use to eager load code and the routes file uses top level method calls, it crashes the application.

If we wrap these routes with `Lotus::Router.define`, the block doesn't get yielded but just returned to the caller as it is.

Usually the receiver of this block is `Lotus::Router#initialize`, which finally evaluates the block.

/cc @lotus/core-team @weppos

### DIFF
--- a/lib/lotus/router.rb
+++ b/lib/lotus/router.rb
@@ -71,6 +71,39 @@ module Lotus
   #
   #   # All the requests starting with "/api" will be forwarded to Api::App
   class Router
+    # Returns the given block as it is.
+    #
+    # When Lotus::Router is used as a standalone gem and the routes are defined
+    # into a configuration file, some systems could raise an exception.
+    #
+    # Imagine the following file into a Ruby on Rails application:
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
It seems perfect ;)

jodosha:
I also manually tested it into a Rails app in production mode and it works great! /cc @weppos

AlfonsoUceda:
@jodosha maybe this can be happens in model for mappings

jodosha:
@AlfonsoUceda Thanks for reminding me. Merging this! :+1:

weppos:
:heart:

PR #78, opened 2015-12-19
Add option for display header text for routes inspector

davydovanton:
## Why this is needful
This option will provide better formatting for `lotus routes` command, for example:
```
~/work/lotus_server » lotus routes
                Name Method   Path                      Action
                     POST     /api/action               Api::Controllers::Files::Action
```

/cc @jodosha

### DIFF
--- a/lib/lotus/routing/routes_inspector.rb
+++ b/lib/lotus/routing/routes_inspector.rb
@@ -18,6 +18,29 @@ module Lotus
       # @api private
       HTTP_METHODS_SEPARATOR = ', '.freeze
 
+      # Default inspector header hash values
+      #
+      # @since 0.5.0
+      # @api private
+      INSPECTOR_HEADER_HASH =  Hash[
+        name:     'Name',
+        methods:  'Method',
+        path:     'Path',
+        endpoint: 'Action'
+      ].freeze
+
+      # Default inspector header name values
+      #
+      # @since 0.5.0
+      # @api private
+      INSPECTOR_HEADER_NAME = 'Name'.freeze
+
+      # Empty line string
+      #
+      # @since 0.5.0
+      # @api private
+      EMPTY_LINE = "\n".freeze
+
       # Instantiate a new inspector
       #
       # @return [Lotus::Routing::RoutesInspector] the new instance
@@ -50,8 +73,10 @@ module Lotus
       #   end
       #
       #   puts router.inspector.to_s
-      #     # =>        GET, HEAD  /                        Home::Index
-      #          login  GET, HEAD  /login                   Sessions::New
+      #     # =>   Name Method     Path                     Action
+      #
+      #                 GET, HEAD  /                        Home::Index
+      #           login GET, HEAD  /login                   Sessions::New
       #                 POST       /login                   Sessions::Create
       #          logout GET, HEAD  /logout                  Sessions::Destroy
       #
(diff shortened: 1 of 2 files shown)


jodosha:
@davydovanton This is good for clarity :+1: 

I'd add a line of separation between the headers and the routes.

lib/lotus/routing/routes_inspector.rb
```diff
@@ -18,6 +18,17 @@ class RoutesInspector
...
+      # Default inspector header hash values
+      #
+      # @since 0.5.0
+      # @api private
+      INSPECTOR_HEADER_HASH =  Hash[
+        name:     'Name'.freeze,
+        methods:  'Method'.freeze,
+        path:     'Path'.freeze,
+        endpoint: 'Action'.freeze
+      ]
```
jodosha: @davydovanton Can you please freeze the hash instead of the single values?

lib/lotus/routing/routes_inspector.rb
```diff
@@ -32,6 +43,7 @@ def initialize(routes)
       #
       # @param formatter [String] the optional formatter for a route
       # @param base_path [String] the base path of a nested route
+      # @param header [Boolean] display header test (default false)
```
jodosha: Please remove this boolean. They are always a code smell. If we think that the header line is useful (as I think it is), there is no need of this flag. All the inspections will have this needed information by default.

lib/lotus/routing/routes_inspector.rb
```diff
@@ -119,6 +148,10 @@ def to_s(formatter = FORMATTER, base_path = nil)
           end
         end
 
+        if header && !result.include?('Name'.freeze)
```
jodosha: Inline freeze is a 2.2+ feature, we're still supporting 2.1. Please add a constant.

lib/lotus/routing/routes_inspector.rb
davydovanton: nice idea, I like this :+1:

davydovanton:
@jodosha done. If you want I can squash the commits to one

joneslee85:
nice one

davydovanton:
thanks @joneslee85! :blush:

jodosha:
Please squash them!

http://lucaguidi.com

davydovanton:
@jodosha done :tada:

jodosha:
@davydovanton :v:

davydovanton:
Thanks! :star2: 

_**P.S.:** @jodosha I want to add search pattern (like in https://github.com/rails/rails/issues/18902) what do you think it's good idea?_

jodosha:
@davydovanton No thanks :smile: 

When in doubt, always rely on UNIX.

davydovanton:
:trollface:

jodosha:
@davydovanton No, I wasn't trolling. Sorry if you've got that feeling. Let me better explain.

Our inspector relies on Ruby's way to print on stdout. That is compliant with UNIX pipes. So we don't have to worry how the output of our command is used by other programs like `grep`.

If we include that `-g` option in our command. We need to take care that it works similarly to `grep` and that the output is compliant with UNIX pipes. This increases the maintenance load for us. Think of potential bugs because we don't respect UNIX standards.

Because `grep` is around [way before I was born](https://en.wikipedia.org/wiki/Grep), and it works great, I don't see the reason to reproduce it.

The less features you add, the less you have to worry about.

davydovanton:
wow, thanks for explanation! I agree with you and I think the Unix way is missing in rails.

jodosha:
A further clarification on what stated above. It's possible to make this feature fully compliant with UNIX. That `-g <routename>` can be used to filter the routes internally and only to print those who are matching the input.

Pseudo code:

```ruby
routes = ['books', 'book']
input   = 'books'

result = routes.find_all do |route|
  route.match /#{ input }/
end

puts result
```

However, `grep` already does that.

Issue #82, opened 2016-01-12
Print extra line at the end of routes inspection


jodosha:
When multiple applications are present in a Lotus project, the output of `lotus routes` is cluttered because the lack of extra blank line between groups of routes:

Steps to reproduce:

```shell
➜ lotus new bookshelf && cd bookshelf && bundle
➜ bundle exec lotus generate app admin
➜ bundle exec lotus generate action web books#index --url=/books
➜ bundle exec lotus generate action admin users#index --url=/users
➜ bundle exec lotus routes
                Name Method     Path                           Action

                     GET, HEAD  /admin/users                   Admin::Controllers::Users::Index
                Name Method     Path                           Action

                     GET, HEAD  /books                         Web::Controllers::Books::Index
```

Expected would be:

```shell
➜ bundle exec lotus routes
                Name Method     Path                           Action
                     GET, HEAD  /admin/users                   Admin::Controllers::Users::Index

                Name Method     Path                           Action
                     GET, HEAD  /books                         Web::Controllers::Books::Index
```

---
Ref #78
/cc @davydovanton

davydovanton:
good idea I think :+1: 
but may be is better to create smth like this:
```
                Name Method     Path                           Action
         --------------------------------------------------------------------------------------
                     GET, HEAD  /admin/users                   Admin::Controllers::Users::Index

                Name Method     Path                           Action
         --------------------------------------------------------------------------------------
                     GET, HEAD  /books                         Web::Controllers::Books::Index
```

jodosha:
@davydovanton With that lines they look grouped like illustrated by the following screenshot:

![screen shot 2016-01-12 at 13 51 14](https://cloud.githubusercontent.com/assets/5089/12263991/ca284530-b933-11e5-8fc7-78586f5ee9ec.png)

davydovanton:
so, @jodosha may be is better create smth like this?
```
                Name Method     Path                           Action

                     GET, HEAD  /admin/users                   Admin::Controllers::Users::Index
                     GET, HEAD  /admin/users                   Admin::Controllers::Users::Index

         --------------------------------------------------------------------------------------

                Name Method     Path                           Action

                     GET, HEAD  /books                         Web::Controllers::Books::Index
                     GET, HEAD  /books                         Web::Controllers::Books::Index
                     GET, HEAD  /books                         Web::Controllers::Books::Index
```

jodosha:
@davydovanton It's way better. It will be probably appear like this instead:

```
[...]

PR #83, opened 2016-01-15
Show header for 'lotus route' command only one time

davydovanton:
This PR do two things:

Firstly, it replaces conditional with polymorphism in `RoutesInspector#to_s` and it closes one TODO task.
Secondly, it fixes displaying header for `lotus routes` command(#82). Now this header shows only one time.

### DIFF
--- a/lib/lotus/routing/routes_inspector.rb
+++ b/lib/lotus/routing/routes_inspector.rb
@@ -135,7 +135,25 @@ module Lotus
       #          | GET, HEAD |      | /api/second_mount/comments | Comments::Index |
       def to_s(formatter = FORMATTER, base_path = nil)
         base_path = Utils::PathPrefix.new(base_path)
-        result    = ''
+
+        inspect_routes(formatter, base_path)
+          .insert(0, formatter % INSPECTOR_HEADER_HASH + EMPTY_LINE)
+      end
+
+      # Returns a string representation of routes
+      #
+      # @param formatter [String] the template for the output
+      # @param base_path [Lotus::Utils::PathPrefix] the base path
+      #
+      # @return [String] serialized routes from router
+      #
+      # @since 0.2.0
+      # @api private
+      #
+      # @see Lotus::Routing::RoutesInspector#FORMATTER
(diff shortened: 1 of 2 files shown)


jodosha:
@davydovanton Hey, thanks for this PR. It changes too much for a "print an extra line for the inspector" fix. What I don't like is that it increases the API of Router with an extra method for inspection.

Can we just have this fix (with a test) for now? Thank you very much.

davydovanton:
@jodosha the base problem in recursion calling `to_s` method. Therefore If you add some title to `#to_s` this method displays this title repeatedly.
I can revert refactoring changes and create private method only for displaying raw text about routes (without any titles or headers).
If you not oppose to such a solution, I can fix it.

P.S.: so, what about refactoring? I can open separate PR only with this.

jodosha:
@davydovanton Please just fix that problem and leave the rest as it is. It's not important, really. We want to focus on new features, rather than on these small details :wink:

davydovanton:
@jodosha I reverted refactoring changes and I replaced recursive call to separate method. Now when you call `lotus routes` command you'll see something like this:
```
                Name Method     Path                           Action

                     GET, HEAD  /fakeroute1                    Fake::Index
                     GET, HEAD  /fakeroute3                    Fake::Index
                     GET, HEAD  /fakeroute2                    Fake::Index
                     GET, HEAD  /fakeroute4                    Fake::Index
                     GET, HEAD  /api/posts                     Posts::Index
                     GET, HEAD  /api/posts_all                 Posts::Index
                     GET, HEAD  /api/second_mount/comments     Comments::Index
                     GET, HEAD  /api/second_mount/comments1    Comments::Index
                     GET, HEAD  /api/second_mount/comments2    Comments::Index
```

jodosha:
@davydovanton Thanks for this update, but there aren't tests here to prove this change. Can you please add them? Thank you! :+1:

davydovanton:
@jodosha done :)

jodosha:
@davydovanton Thank you :+1:

davydovanton:
Thanks for review :+1:

PR #86, opened 2016-01-25
Fix body parsers for non Hash requests

### LINKED ISSUE #85: The router crashes when the body is a JSON array
I was able to reproduce the error [here](https://github.com/weppos/hanami-router/commit/5ea3b15b96094c5d9da903648b56db69219ff95c).

```
  1) Error:
Body parsing#test_0002_is successful (JSON) with an array:
TypeError: no implicit conversion of Array into Hash
    /Users/weppos/Mirrors/hanami--router/lib/hanami/routing/parsers.rb:49:in `merge!'
    /Users/weppos/Mirrors/hanami--router/lib/hanami/routing/parsers.rb:49:in `block in _redefine_call'
    /Users/weppos/Mirrors/hanami--router/lib/hanami/routing/http_router.rb:135:in `raw_call'
    /Users/weppos/.rvm/gems/ruby-2.3.0/gems/http_router-0.11.2/lib/http_router.rb:142:in `call'
    /Users/weppos/Mirrors/hanami--router/lib/hanami/router.rb:931:in `call'
    /Users/weppos/.rvm/gems/ruby-2.3.0/gems/rack-1.6.4/lib/rack/lint.rb:49:in `_call'
    /Users/weppos/.rvm/gems/ruby-2.3.0/gems/rack-1.6.4/lib/rack/lint.rb:37:in `call'
    /Users/weppos/.rvm/gems/ruby-2.3.0/gems/rack-1.6.4/lib/rack/mock.rb:74:in `request'
    /Users/weppos/.rvm/gems/ruby-2.3.0/gems/rack-1.6.4/lib/rack/mock.rb:59:in `patch'
    /Users/weppos/Mirrors/hanami--router/test/integration/body_parsing_test.rb:27:in `block (2 levels) in <top (required)>'
```

```ruby
  it 'is successful (JSON) with an array' do
    body     = StringIO.new( %(["alpha", "beta"]).encode(Encoding::ASCII_8BIT) )
    response = @app.patch('/books/23', 'CONTENT_TYPE' => 'application/json', 'rack.input' => body, lint: true)

    response.status.must_equal 200
    response.body.must_equal %(["alpha", "beta"])
  end
```

This is an example of a request

```
$ curl http://example.com/path -X POST -d '["alpha", "beta"]' -H "Content-Type: application/json"
```

Of course, the tricky part is how this case should be handled in the `params` as it implies the `params` can also be an `Array`, instead of always a `Hash`.

It's possible to use a constant key, for example `_`, which is also the same approach Rails is following.

```
{"_json"=>["ns1.example.com", "ns2.example.com"], "controller"=>"something", "action"=>"create", "something"=>{"_json"=>["ns1.example.com", "ns2.example.com"]}}
```

In this case, Rails uses the `_json` key.

FYI, Lotus crashes at this point, because `merge!` can't handle an input which is not a `Hash`.

```ruby
          env[ROUTER_PARAMS].merge!(
            @parsers[
              media_type(env)
            ].parse(body)
          )
```

jodosha:
Closes #85 

---

Thanks to @weppos for spotting the problem and reproducing it. :+1: 

/cc @hanami/core-team

### DIFF
--- a/lib/hanami/routing/parsers.rb
+++ b/lib/hanami/routing/parsers.rb
@@ -8,6 +8,7 @@ module Hanami
 
       RACK_INPUT    = 'rack.input'.freeze
       ROUTER_PARAMS = 'router.params'.freeze
+      FALLBACK_KEY  = '_'.freeze
 
       def initialize(parsers)
         @parsers = prepare(parsers)
@@ -44,15 +45,26 @@ module Hanami
           return env if body.empty?
 
           env[RACK_INPUT].rewind    # somebody might try to read this stream
-          env[ROUTER_PARAMS] ||= {} # prepare params
 
+          env[ROUTER_PARAMS] ||= {} # prepare params
           env[ROUTER_PARAMS].merge!(
-            @parsers[
-              media_type(env)
-            ].parse(body)
+            _parse(env, body)
           )
 
           env
+        end
+      end
+
+      def _parse(env, body)
+        result = @parsers[
+          media_type(env)
+        ].parse(body)
+
+        case result
+        when Hash
+          result
+        else
+          {FALLBACK_KEY => result}
         end
       end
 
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
:+1:

Issue #87, opened 2016-01-25
Setting SCRIPT_NAME environment variable when mounting Rack apps


stravid:
When I setup [Sidekiq with Hanami](http://www.strauss.io/blog/2016-use-sidekiq-with-hanami.html) I ran into an issue when mounting the Sidekiq web interface within an applications route or the container. All URLs for links and assets in Sidekiq were broken because they pointed to the root of my Hanami application even if I mounted it at `/sidekiq` for example.

Turns out [Sidekiq uses `SCRIPT_NAME` to build its URLs](https://github.com/mperham/sidekiq/blob/f8baa8af37274cacd3ef0437e9cede5ad5ecbe0d/lib/sidekiq/web_helpers.rb#L137) and it seems that Hanami only sets it for Hanami apps and not Rack apps in general.

Reading the [Rack specification](http://www.rubydoc.info/github/rack/rack/file/SPEC) I think Hanami should set `SCRIPT_NAME` accordingly. `Rack::URLMap` [does it too for example](https://github.com/rack/rack/blob/028438ffffd95ce1f6197d38c04fa5ea6a034a85/lib/rack/urlmap.rb#L63).

> The initial portion of the request URL's “path” that corresponds to the application object, so that the application knows its virtual “location”. This may be an empty string, if the application corresponds to the “root” of the server.

What do you think?

jodosha:
There is a PR by @cyphactor that should fix this problem directly into `http_router` (See https://github.com/joshbuddy/http_router/pull/43).

If there isn't a quick merge and release of `http_router`, we can act now and port that patch here. To solve this kind of problems that we had in the past `Hanami::Router` wraps `Hanami::Routing::HttpRouter` which is a subclass of `HttpRouter` (from `http_router` gem).

@cyphactor It would be really helpful if you can port it here. Thank you very much!

cyphactor:
I still haven't seen any movement on the http_router PR (https://github.com/joshbuddy/http_router/pull/43) sadly. Hence, I don't expect there to be any movement in terms of getting it merged in and released any time soon.

@jodosha I will try to port it to hanami-router today and get a PR up.

jodosha:
Closed by #91 Thanks to @cyphactor :clap:

jamesdabbs:
Hey folks. I'm mucking around trying to get Hanami running on Rack 2 and ran into an issue with one of these tests that I thought would be good to bring up. Here's what I found -

[...]

PR #90, opened 2016-02-09
Fix empty SCRIPT_NAME with partial match route bug

cyphactor:
Why you made the change:

I made this change so that SCRIPT_NAME and PATH_INFO would be set
appropriately in the scenario where a partial match route is used.

There is an issue with the current version of rewrite_partial_path_info
in HttpRouter where it does not set SCRIPT_NAME appropriately. When the
request environments PATH_INFO is set, it causes the
request.rack_request.path_info helper return the updated version, not
the original. This is requst.rack_request.path_info eventually just
reads the PATH_INFO value out of the request environment. As a side
effect of this, the original code would always set the SCRIPT_NAME to ""
because it would be slicing from request.rack_request.path_info[0, 0] in
actuality.

This change resolves this issue by first temporarily storing a dup of
the original path_info and using that to determine the end index for the
SCRIPT_NAME. This change also handles setting SCRIPT_NAME correctly for
the somewhat unique but common case of matching the partial match
exactly.

This change also includes a fix to handle maintaining URI encoded values
in PATH_INFO and SCRIPT_INFO. Currently, there is a problem where
because the method uses the request.path helper HttpRouter provides,
which URI.decodes the string, it screws up both the PATH_INFO and
SCRIPT_NAME. Therefore, I URI.encode it back appropriately before any
offsetting, or setting is done.

I have also added tests for each of these cases as I couldn't find any
tests covering this before.

This fixes issue #87.

### DIFF
--- a/lib/hanami/routing/http_router.rb
+++ b/lib/hanami/routing/http_router.rb
@@ -164,6 +164,18 @@ module Hanami
         env[SCRIPT_NAME] = @prefix.join(env[SCRIPT_NAME])
       end
 
+      # @api private
+      def rewrite_partial_path_info(env, request)
+        path_info_before = request.rack_request.path_info.dup
+        if request.path.empty?
+          env['PATH_INFO'] = "/"
+          env['SCRIPT_NAME'] += path_info_before
+        else
+          env['PATH_INFO'] = "/#{URI.encode(request.path.join('/'))}"
+          env['SCRIPT_NAME'] += path_info_before[0, path_info_before.size - env['PATH_INFO'].size]
+        end
+      end
+
       private
 
       def _rescue_url_recognition
--- a/test/routing/http_router_test.rb
+++ b/test/routing/http_router_test.rb
@@ -24,4 +24,60 @@ describe Hanami::Routing::HttpRouter do
       env['SCRIPT_NAME'].to_s.must_equal '/post'
     end
   end
+
+  describe '#rewrite_partial_path_info' do
+    describe 'when from partial match' do
+      it 'sets PATH_INFO correctly' do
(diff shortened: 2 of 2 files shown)


cyphactor:
@jodosha you may have noticed that this hasn't passed CI. That is because this change introduces a failure in an existing test and I wanted to discuss it.

If I am not mistaken, the failure is actually happening because the changes I made are correct and it is now showing that the `HttpRouter`'s splitting logic based on what it matches in this particular test case is wrong.

What I mean by this is that in `HttpRouter` the `request.path` override they provide in the `HttpRouter::Request` has already gone through the splitting logic by the time the request reaches the `rewrite_partial_path_info` method that I corrected. This is important because within the scope of this method you do not technically have enough knowledge to properly decide which portion of the full path belongs to the `SCRIPT_NAME` and which portion belongs to the `PATH_INFO`. So, it seems that test was passing before as a side effect of having `SCRIPT_NAME` and `PATH_INFO` being set incorrectly.

Please verify that you agree with me or not. I am really looking for a second pair of eyes to verify this before diving into the splitting logic in `HttpRouter`. From a brief glance it looks to be nightmarish.

cyphactor:
I am closing this PR because further work is need either in removing the prefix concept or figuring out how it should play properly with this fix because the prefix change seems to be a workaround for this issue being broken.

The following will help provide some context.

https://github.com/hanami/router/commit/b6d1db11fdc3865754e44239c4f6e81524109173

https://github.com/hanami/router/commit/b6d1db11fdc3865754e44239c4f6e81524109173#diff-2930410f18b220852ff26d55d8af37d7R157

PR #91, opened 2016-02-10
Fix empty script name with partial match route bug

cyphactor:
Why you made the change:

I made this change so that SCRIPT_NAME and PATH_INFO would be set
appropriately in the scenario where a partial match route is used.

There is an issue with the current version of rewrite_partial_path_info
in HttpRouter where it does not set SCRIPT_NAME appropriately. When the
request environments PATH_INFO is set, it causes the
request.rack_request.path_info helper return the updated version, not
the original. This is requst.rack_request.path_info eventually just
reads the PATH_INFO value out of the request environment. As a side
effect of this, the original code would always set the SCRIPT_NAME to ""
because it would be slicing from request.rack_request.path_info[0, 0] in
actuality.

This change resolves this issue by first temporarily storing a dup of
the original path_info and using that to determine the end index for the
SCRIPT_NAME. This change also handles setting SCRIPT_NAME correctly for
the somewhat unique but common case of matching the partial match
exactly.

This change also includes a fix to handle maintaining URI encoded values
in PATH_INFO and SCRIPT_INFO. Currently, there is a problem where
because the method uses the request.path helper HttpRouter provides,
which URI.decodes the string, it screws up both the PATH_INFO and
SCRIPT_NAME. Therefore, I URI.encode it back appropriately before any
offsetting, or setting is done.

I have also added tests for each of these cases as I couldn't find any
tests covering this before.

This fixes issue #87.

### DIFF
--- a/lib/hanami/routing/http_router.rb
+++ b/lib/hanami/routing/http_router.rb
@@ -37,6 +37,11 @@ module Hanami
       # @api private
       SCRIPT_NAME = 'SCRIPT_NAME'.freeze
 
+      # Path info - rack environment variable
+      #
+      # @api private
+      PATH_INFO = 'PATH_INFO'.freeze
+
       # @since 0.5.0
       # @api private
       attr_reader :namespace
@@ -158,21 +163,14 @@ module Hanami
       end
 
       # @api private
-      # @since 0.5.0
-      def rewrite_path_info(env, request)
-        super
-        env[SCRIPT_NAME] = @prefix.join(env[SCRIPT_NAME])
-      end
-
-      # @api private
       def rewrite_partial_path_info(env, request)
         path_info_before = request.rack_request.path_info.dup
         if request.path.empty?
-          env['PATH_INFO'] = "/"
-          env['SCRIPT_NAME'] += path_info_before
+          env[PATH_INFO] = "/"
+          env[SCRIPT_NAME] += path_info_before
         else
-          env['PATH_INFO'] = "/#{URI.encode(request.path.join('/'))}"
-          env['SCRIPT_NAME'] += path_info_before[0, path_info_before.size - env['PATH_INFO'].size]
+          env[PATH_INFO] = "/#{URI.encode(request.path.join('/'))}"
+          env[SCRIPT_NAME] += path_info_before[0, path_info_before.size - env[PATH_INFO].size]
         end
       end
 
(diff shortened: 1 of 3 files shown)


cyphactor:
@jodosha I went through this in depth today. We are good to go now I believe. Couldn't get rid of `prefix` completely because it is still needed for making the path & url helpers work correctly.

test/routing/http_router_test.rb
```diff
@@ -24,4 +24,60 @@ def rack_request
       env['SCRIPT_NAME'].to_s.must_equal '/post'
     end
   end
+
+  describe '#rewrite_partial_path_info' do
+    describe 'when from partial match' do
+      it 'sets PATH_INFO correctly' do
+        request_env = nil
+        router = Hanami::Routing::HttpRouter.new
+        router.add("/sidekiq*").to { |env| request_env = env; [200, {}, []] }
```
joneslee85: you can group common codes to a before block

joneslee85:
except above comments, it is :+1:

test/routing/http_router_test.rb
jodosha: @cyphactor :+1: to @joneslee85 comment

cyphactor:
Thx @joneslee85 and @jodosha for the review. I just pushed up a commit to add the before block and collapse the common portions of those tests into the before block.

lib/hanami/routing/http_router.rb
```diff
@@ -158,10 +163,15 @@ def no_response(request, env)
...
-        super
-        env[SCRIPT_NAME] = @prefix.join(env[SCRIPT_NAME])
+      def rewrite_partial_path_info(env, request)
+        path_info_before = request.rack_request.path_info.dup
+        if request.path.empty?
+          env[PATH_INFO] = "/"
+          env[SCRIPT_NAME] += path_info_before
+        else
+          env[PATH_INFO] = "/#{URI.encode(request.path.join('/'))}"
+          env[SCRIPT_NAME] += path_info_before[0, path_info_before.size - env[PATH_INFO].size]
```
jodosha: @cyphactor Shouldn't we use `bytesize` instead of `size`? It's safer for Unicode chars.

cyphactor:
@jodosha pushed up change to `bytesize`.

joneslee85:
:+1:

jodosha:
@cyphactor I've merged it and did some cleanups (see https://github.com/hanami/router/commit/c6f527280945f7a973bf4c55c6b02185247478cf). Thank you very much for this fix :+1:

cyphactor:
@jodosha no problem. Thanks for the cleanups. Next time I will try and keep those in mind.

joneslee85:
@cyphactor thanks for the PR

PR #93, opened 2016-02-18
Make hanami routes work with nested non-hanami apps

### LINKED ISSUE #92: hanami routes fails when Sinatra app mounted in container
It seems in the scenario where you have a Container architecture hanami app and have a Sinatra app mounted it blows up with an exception because the inspector assumes a `Hanami::Router` instance. I discovered this because I was mounting Sidekiq's web interface into my hanami container application as follows:

```ruby
Hanami::Container.configure do
  mount Sidekiq::Web, at: '/sidekiq'
end

```

The following is the traceback. I believe it would potentially be as simple as checking if the router responds to `inspector` and if it doesn't ignore it.

```text
/Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-router-0.6.2/lib/hanami/routing/routes_inspector.rb:213:in `inspect_router': undefined method `inspector' for #<Hash:0x007ff13b607780> (NoMethodError)
Did you mean?  inspect
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-router-0.6.2/lib/hanami/routing/routes_inspector.rb:163:in `block in inspect_routes'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-router-0.6.2/lib/hanami/routing/routes_inspector.rb:161:in `each'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-router-0.6.2/lib/hanami/routing/routes_inspector.rb:161:in `inspect_routes'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-router-0.6.2/lib/hanami/routing/routes_inspector.rb:139:in `to_s'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-0.7.2/lib/hanami/commands/routes.rb:25:in `start'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-0.7.2/lib/hanami/cli.rb:103:in `routes'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
	from /Users/adeponte/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/hanami-0.7.2/bin/hanami:5:in `<top (required)>'
	from /Users/adeponte/.rbenv/versions/2.3.0/bin/hanami:23:in `load'
	from /Users/adeponte/.rbenv/versions/2.3.0/bin/hanami:23:in `<main>'

```

cyphactor:
Why you made the change:

I did this so that non-hanami applications like Sinatra, apps, etc. when
mounted will still allow the `hanami routes` command line interface to
work.

As discussed in issue #92 there is an issue where when a non-hanami
application is mounted that responds to the `routes` method that doesn't
return a `Hanami::Router` it throws an exception. This change fixes that
by identifying non-hanami nested routers properly rather than just doing
it based the application responding to the `routes` method.

---

Closes #92

### DIFF
--- a/lib/hanami/routing/route.rb
+++ b/lib/hanami/routing/route.rb
@@ -45,13 +45,12 @@ module Hanami
       end
 
       # Introspect the given route to understand if there is a wrapped
-      # Hanami::Router
+      # router that has an inspector
       #
       # @since 0.2.0
       # @api private
       def nested_router
-        dest.routes if dest.respond_to?(:routes) &&
-          dest.routes.is_a?(Hanami::Router)
+        dest.routes if dest.respond_to?(:routes) && dest.routes.respond_to?(:inspector)
       end
 
       private
--- a/lib/hanami/routing/route.rb
+++ b/lib/hanami/routing/route.rb
@@ -50,7 +50,8 @@ module Hanami
       # @since 0.2.0
       # @api private
       def nested_router
-        dest.routes if dest.respond_to?(:routes)
+        dest.routes if dest.respond_to?(:routes) &&
+          dest.routes.is_a?(Hanami::Router)
       end
 
       private
--- a/test/routing/routes_inspector_test.rb
+++ b/test/routing/routes_inspector_test.rb
@@ -224,10 +224,18 @@ describe Hanami::Routing::RoutesInspector do
           mount inner_router, at: '/second_mount'
         }
 
+        nested_non_hanami_router = Object.new
+        def nested_non_hanami_router.call(env)
+        end
+        def nested_non_hanami_router.routes
+          return {}
+        end
+
         @router = Hanami::Router.new do
           get '/fakeroute', to: 'fake#index'
           mount AdminHanamiApp,  at: '/admin'
           mount nested_router,  at: '/api'
+          mount nested_non_hanami_router, at: '/foo'
           mount RackMiddleware, at: '/class'
           mount RackMiddlewareInstanceMethod,     at: '/instance_from_class'
           mount RackMiddlewareInstanceMethod.new, at: '/instance'
@@ -243,7 +251,8 @@ describe Hanami::Routing::RoutesInspector do
           %(| GET, HEAD |  | /api/second_mount/comments | Comments::Index |),
           %(|  |  | /class | RackMiddleware |),
           %(|  |  | /instance_from_class | #<RackMiddlewareInstanceMethod> |),
-          %(|  |  | /instance | #<RackMiddlewareInstanceMethod> |)
+          %(|  |  | /instance | #<RackMiddlewareInstanceMethod> |),
+          %(|  |  | /foo | #<Object> |)
         ]
 
         actual = @router.inspector.to_s(formatter)


lib/hanami/routing/route.rb
```diff
@@ -50,7 +50,8 @@ def generate(resolver, options = {}, &endpoint)
       # @since 0.2.0
       # @api private
       def nested_router
-        dest.routes if dest.respond_to?(:routes)
+        dest.routes if dest.respond_to?(:routes) &&
```
cyphactor: One thing to note here is that the `dest.respond_to?(:routes) &&` is possibly no longer needed as it was a less ideal way of identifying Hanami::Router instances before. If you agree this is no longer needed I will remove it.

jodosha: @cyphactor At the next line there is `dest.routes` invocation. If we remove this `respond_to?` check, it can crash for object that don't respond to `#routes`. :wink:

lib/hanami/routing/route.rb
```diff
@@ -50,7 +50,8 @@ def generate(resolver, options = {}, &endpoint)
...
+          dest.routes.is_a?(Hanami::Router)
```
jodosha: @cyphactor What if instead of checking for type, we still check for behavior? In a language like Ruby, this is always an advantage for openness.

For instance:

```ruby
def nested_router
  if dest.respond_to?(:routes) && dest.routes.respond_to?(:inspector)
    dest.routes
  end
end
```

Do you think it can still work?

cyphactor: After digging through the code base a bit more I believe we can do the more duck type style behavior checking with a check for responding to both `routes` and `routes.inspector` because it seems the next call on the `inspector` is `to_s` which everything has thanks to `Object`.

It is important to note that in scenarios where the nested route object does respond to `inspector` but the `to_s` of that object is different than that of `Routing::RoutesInspector` it may potentially make the `hanami routes` output look strange or screwed up. However, it does leave things more open and would actually allow other types of routes that are inspectable to be able to operate within `hanami routes`.

I will push the duck-type styled version up shortly.

jodosha: > it may potentially make the hanami routes output look strange or screwed up.

We don't expect to support what we don't know, but to make it open for others to provide a routes good inspection. If you mount a non "inspectable" Rack app, it's not a big deal if the routes printing isn't pretty.

cyphactor: Cool. I think we should be good to go then.

jodosha:
I tried and it works:

```shell
➜ be hanami routes
                Name Method     Path                           Action

                                /sidekiq                       Sidekiq::Web
```

jodosha:
@cyphactor Thank you very much, merging this! :+1:

PR #98, opened 2016-03-07
Add root route

### LINKED ISSUE #97: Automatic root_path helper?
Making a `hanami` app and defining a route at `'/'`, I expected I'd have access to a `routes.root_path` helper, but I had to add `as: :root` in order to get it.

This could that I'm accustomed to Rails, and that `hanami` prefers to be explicit, but automatically adding a `root` path for `'/'` seems like a nice feature :)

cllns:
Closes #97.

The guides will need to be updated, which I can do after this is merged. 

Do you think we should add support for missing the `to:`? e.g. `root "welcome#index"`
[Rails has this shortcut](http://guides.rubyonrails.org/routing.html#using-root) and it's nice.

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -491,6 +491,37 @@ module Hanami
       @router.trace(path, options, &blk)
     end
 
+    # Defines a root route (a GET route for '/')
+    #
+    # @param options [Hash] the options to customize the route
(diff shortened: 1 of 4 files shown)


jodosha:
@cllns This looks good. I merged it. :+1:

PR #104, opened 2016-04-11
Match against host

nicolaracco:
This way it's possible to mount different containers on different domains. E.g.:

```
Hanami::Container.configure do
  mount Admin::Application, at: '/', host: 'admin.hanami.dev'
  mount Web::Application, at: '/'
end
```

### DIFF
--- a/lib/hanami/routing/http_router.rb
+++ b/lib/hanami/routing/http_router.rb
@@ -143,7 +143,7 @@ module Hanami
       # @since 0.1.1
       # @api private
       def mount(app, options)
-        add("#{ options.fetch(:at) }*").to(
+        add("#{ options.fetch(:at) }*", host: options[:host]).to(
           @resolver.resolve(to: app)
         )
       end
--- a/test/integration/router_container_test.rb
+++ b/test/integration/router_container_test.rb
@@ -13,4 +13,17 @@ describe 'Router wrapper as container' do
     response = @app.get('/back/home', lint: true)
     response.body.must_equal 'back'
   end
+
+  it 'matches against host' do
+    @router_container = Hanami::Router.new(scheme: 'https', host: 'hanami.test', port: 443) do
+      mount Front::App, at: '/front', host: 'www.hanami.test'
+      mount Back::App, at: '/front'
+    end
+
+    @app = Rack::MockRequest.new(@router_container)
+    response = @app.get('https://www.hanami.test/front/home', lint: true)
+    response.body.must_equal 'front'
+    response = @app.get('/front/home', lint: true)
+    response.body.must_equal 'back'
+  end
 end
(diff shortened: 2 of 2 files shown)


lib/hanami/routing/http_router.rb
```diff
@@ -143,7 +143,7 @@ def options(path, options = {}, &blk)
       # @since 0.1.1
       # @api private
       def mount(app, options)
-        add("#{ options.fetch(:at) }*").to(
+        add("#{ options.fetch(:at) }*", host: options[:host]).to(
```
jodosha: @nicolaracco Using `:host` option both as `Router#initialize` and as a `#mount` option can be confusing. What about `:subdomain` instead?

test/integration/router_container_test.rb
```diff
@@ -13,4 +13,17 @@
     response = @app.get('/back/home', lint: true)
     response.body.must_equal 'back'
   end
+
+  it 'matches against host' do
+    @router_container = Hanami::Router.new(scheme: 'https', host: 'hanami.test', port: 443) do
+      mount Front::App, at: '/front', host: 'www.hanami.test'
```
jodosha: @nicolaracco What happens if a developer only uses the subdomain? Eg. `host: 'www'`.

test/integration/router_container_test.rb
```diff
@@ -13,4 +13,17 @@
...
+      mount Back::App, at: '/front'
+    end
+
+    @app = Rack::MockRequest.new(@router_container)
+    response = @app.get('https://www.hanami.test/front/home', lint: true)
+    response.body.must_equal 'front'
+    response = @app.get('/front/home', lint: true)
```
jodosha: @nicolaracco Shall we test this like this? `@app.get('https:/hanami.test/front/home', lint: true)`

jodosha:
@nicolaracco Thanks for this PR. This feature was missing and people kept asking about it from time to time. I left a few inline comments.

test/integration/router_container_test.rb
nicolaracco: The main problem is that this `host` parameter actually matches the exact host. It's not clearly documented in the http_router gem but I found it looking at the tests. Specifying the host parameter will actually call the [`Node#add_host`](https://github.com/joshbuddy/http_router/blob/master/lib/http_router/node.rb#L37) method that relies on the [`Node::Host`](https://github.com/joshbuddy/http_router/blob/master/lib/http_router/node/host.rb) class. This class doesn't override the `to_code` method (defined on [`Node::AbstractRequestNode`](https://github.com/joshbuddy/http_router/blob/master/lib/http_router/node/abstract_request_node.rb#L20))  that, by default, seems to execute a simple equality comparison.

I was thinking to do a PR on http_router gem in the future, so that the `host` param can work against a regexp.

jodosha: @nicolaracco I see and that is really annoying. I'm gonna merge this, but after Hanami 1.0, we should revise this problem.

jodosha:
@nicolaracco Merged, thank you! 🎉

PR #105, opened 2016-04-29
Use MultiJSON to allow usage of different JSON engines

Erol:
What do you guys say to using [Oj](https://github.com/ohler55/oj) for parsing JSON requests?

Here's the result of `benchmarks/json` on my machine. Ruby's JSON parser is 1.74x slower than Oj.

```
Calculating -------------------------------------
                json   981.000  i/100ms
                  oj     1.734k i/100ms
-------------------------------------------------
                json     10.016k (± 3.5%) i/s -     50.031k
                  oj     17.393k (± 9.1%) i/s -     86.700k

Comparison:
                  oj:    17392.7 i/s
                json:    10016.2 i/s - 1.74x slower
```

### DIFF
--- a/hanami-router.gemspec
+++ b/hanami-router.gemspec
@@ -20,10 +20,12 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.2.0'
 
   spec.add_dependency 'http_router',  '~> 0.11'
+  spec.add_dependency 'multi_json',   '~> 1.11.3'
   spec.add_dependency 'hanami-utils', '~> 0.8'
 
-  spec.add_development_dependency 'bundler',   '~> 1.5'
-  spec.add_development_dependency 'minitest',  '~> 5'
-  spec.add_development_dependency 'rake',      '~> 11'
-  spec.add_development_dependency 'rack-test', '~> 0.6'
+  spec.add_development_dependency 'benchmark-ips', '~> 2.3.0'
+  spec.add_development_dependency 'bundler',       '~> 1.5'
+  spec.add_development_dependency 'minitest',      '~> 5'
+  spec.add_development_dependency 'rake',          '~> 11'
+  spec.add_development_dependency 'rack-test',     '~> 0.6'
 end
--- a/benchmarks/json
+++ b/benchmarks/json
@@ -1,0 +1,111 @@
+#!/usr/bin/env ruby -W0
+require 'benchmark/ips'
+require 'json'
+require 'oj'
+
+json = <<-JSON
+[
+    {
+        "fname": "Keana",
+        "lname": "Holmes",
+        "tel": "(788)347-3637",
+        "address": "3652 Non Ct",
+        "city": "Peoria",
+        "state": "RI",
+        "zip": 36792
+    },
+    {
+        "fname": "Marisa",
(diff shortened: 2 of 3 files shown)


davydovanton:
@Erol thanks for this PR 🎉 
What do you think, may be is better to create special option for select JSON parser engine? Something like this:
``` ruby
  config.json_parser Oj
  # or
  config.json_parser JSON
```

/cc @hanami/contributors @hanami/core-team

Erol:
I like that idea! Given https://github.com/hanami/router#json-parsing, what do you guys think about expanding the options to configure the JSON parser:

```
router = Hanami::Router.new(parsers: [json: Oj]) do
  patch '/books/:id', to: endpoint
end
```

davydovanton:
@Erol yep, it's better for me 👍

hanami-router.gemspec
```diff
@@ -20,10 +20,12 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.2.0'
 
   spec.add_dependency 'http_router',  '~> 0.11'
+  spec.add_dependency 'oj',           '~> 2.15.0' unless RUBY_PLATFORM == 'java'
```
Erol: I'm thinking that with the option to specify which JSON parser to use, we can drop the hard dependency on Oj and let application developers include it on their own.

joneslee85:
@Erol thanks for the PR. There is no doubt about the perf gain by using Oj with MRI. The downside is that Hanami tries to be agnostic as much as possible. Supporting different adapter is a good suggestion and personally I 👍 but I don't introduce `oj` as dependency by default nor want it opted in by default for MRI. IMHO it should be opted in by choice. Thought?

Erol:
@joneslee85: I totally agree with you on this. Actually, when I went deeper into the parsing code I realized that it might be easier and cleaner if parsing via `Oj` is done via a separate library, like:

```
require 'hanami/router'
require 'hanami/json/oj' # Or something similar

router = Hanami::Router.new(parsers: [JSON::Oj]) do
  patch '/books/:id', to: endpoint
end
```

Also, I haven't really gone deep in the Hanami framework code yet, but it might be useful if the parsers could be specified like:

```
config.parsers JSON::Oj
```

What do you guys think?

pascalbetz:
Sounds like we are re-building MultiJSON?

Erol:
Oh yeah, it does look like we're reinventing the wheel here... how about introducing `MultiJson` as a dependency and doing this instead?

```ruby
require 'multi_json'

module Hanami
  module Routing
    module Parsing
      class JsonParser < Parser
        def mime_types
          ['application/json', 'application/vnd.api+json']
        end

        def parse(body)
          MultiJson.load(body)
        rescue MultiJson::ParserError => e
          raise BodyParsingError.new(e.message)
        end
      end
    end
  end
end
```

jodosha:
@Erol Thanks for this proposal, but I'm not keen to add yet another dependency. On the other hand, it can be useful to define your own JSON backend. How about this proposal: #107?

Please folks let me know what do you think about it.

pascalbetz:
Like it. Works out of the box, can be changed if required.

Erol:
@jodosha Yeah let's go with #107. Closing this for now.

PR #107, opened 2016-05-03
Support multi_json for body parsers

jodosha:
Closes #105

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,4 +7,5 @@ if !ENV['TRAVIS']
 end
 
 gem 'hanami-utils', '~> 0.8', require: false, github: 'hanami/utils', branch: '0.8.x'
+gem 'multi_json',   '~> 1.0', require: false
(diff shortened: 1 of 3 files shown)


Erol:
I like it! Just not too comfortable with `JSON` masquerading as `MultiJson`, but other than that this works for me.

Erol:
How about this @jodosha? This way our mock `MultiJson` is scoped inside `JsonParser`:

```ruby
begin
  require 'multi_json'
rescue LoadError
  require 'json'
end

module Hanami
  module Routing
    module Parsing
      class JsonParser < Parser
        unless defined? MultiJson
          MultiJson = JSON
          MultiJson::ParseError = JSON::ParserError
        end

        ...
      end
    end
  end
end
```

jodosha:
@Erol Now we're talking! 😉  Your suggestion is better. 👍

Erol:
Sweet! 👍

PR #110, opened 2016-05-26
Fix typo on EndpointResolver doc

Erol:


jodosha:
@Erol Thank you again! 👯

Issue #115, opened 2016-06-27
Unrecognized route causes a NoMethodError


jodosha:
```shell
  1) API::V2 routing POST [REDACTED]
     Failure/Error: router.__send__(:api_from_registry, actual.action) ==

     NoMethodError:
       undefined method `__getobj__' for nil:NilClass
     # /Users/luca/.gem/ruby/2.3.1/gems/hanami-router-0.6.2/lib/hanami/routing/recognized_route.rb:143:in `destination'
     # /Users/luca/.gem/ruby/2.3.1/gems/hanami-router-0.6.2/lib/hanami/routing/recognized_route.rb:104:in `action'
     # [REDACTED]:318:in `block (2 levels) in <module:Matchers>'
     # [REDACTED]:669:in `block (3 levels) in <top (required)>'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler/cli/exec.rb:63:in `load'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler/cli/exec.rb:63:in `kernel_load'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler/cli/exec.rb:24:in `run'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler/cli.rb:304:in `exec'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler/cli.rb:11:in `start'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/exe/bundle:27:in `block in <top (required)>'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler/friendly_errors.rb:98:in `with_friendly_errors'
     # /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/exe/bundle:19:in `<top (required)>'
```

vladimirich:
Can I take this issue?

vladimirich:
@jodosha any tips, how can I reproduce this bug?

vladimirich:
Ok, seems like I got this. You have no routes at all, something like

```
router = Hanami::Router.new do
end

route = router.recognize('/books/123')
route.action
```

jodosha:
@vladimirich It doesn't just happen when the router is blank, but when the requested route is missing.

PR #116, opened 2016-06-29
fix NoMethodError with blank routes

vladimirich:
#115

### DIFF
--- a/lib/hanami/routing/recognized_route.rb
+++ b/lib/hanami/routing/recognized_route.rb
@@ -140,6 +140,7 @@ module Hanami
       #
       # @see Hanami::Routing::Endpoint
       def destination
+        return '' unless @endpoint
         @destination ||= begin
           case k = @endpoint.__getobj__
           when Class
--- a/lib/hanami/routing/recognized_route.rb
+++ b/lib/hanami/routing/recognized_route.rb
@@ -140,7 +140,6 @@ module Hanami
       #
       # @see Hanami::Routing::Endpoint
       def destination
-        return '' unless @endpoint
         @destination ||= begin
           case k = @endpoint.__getobj__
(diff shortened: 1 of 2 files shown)


lib/hanami/routing/recognized_route.rb
```diff
@@ -140,6 +140,7 @@ def routable?
       #
       # @see Hanami::Routing::Endpoint
       def destination
+        return '' unless @endpoint
         @destination ||= begin
```
joneslee85: what do you think about catching exception instead? since we have already got `begin` block here, you can always do `rescue NoMethodError` then return a value

test/recognize_test.rb
```diff
@@ -269,4 +269,21 @@
       end
     end
   end
+
+  describe '#recognize without routes' do
```
joneslee85: this example description to me is duplication of line 279, there should be only 1

lib/hanami/routing/recognized_route.rb
vladimirich: Ok, will fix soon.

test/recognize_test.rb
vladimirich: @joneslee85 , will fix, ty for feedback!

jodosha:
@vladimirich Can you please check why the build fails? Thanks!

jodosha:
Closing in favor of #117 

@vladimirich Thanks for your attempt, but I had to fix myself for a few reasons. We're about to release Hanami v0.8 in a couple of weeks and almost all the gems must be frozen. This affects a product that runs Hanami in production, and I wanted to make sure that it will work smoothly when upgraded.

Thanks for understanding. Please remember that your contributions are always welcome.

PR #118, opened 2016-08-02
Add parsed body to env

Moratorius:
I did this because I want to access the body parsed with the body parsers outside of `Action#params`.  Currently the parsed body attributes are merged with route/query params as `#params`, which makes it impossible to know the source of a parameter.

### DIFF
--- a/lib/hanami/routing/parsers.rb
+++ b/lib/hanami/routing/parsers.rb
@@ -8,6 +8,7 @@ module Hanami
 
       RACK_INPUT    = 'rack.input'.freeze
       ROUTER_PARAMS = 'router.params'.freeze
+      ROUTER_PARSED_BODY = 'router.parsed_body'.freeze
       FALLBACK_KEY  = '_'.freeze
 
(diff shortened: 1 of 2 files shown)


cllns:
Do we need this if https://github.com/hanami/controller/pull/172 is merged? That seems like a better fix than adding this workaround.

Moratorius:
This change has a different goal than solving the stringification of params.  Combining query params and body attributes together to produce a unified parameters object makes it impossible to know the source of a particular "parameter."  I don't have an issue with leaving that behavior intact as `#params`, but I do want the choice of accessing the result of the body_parsers parse directly when I want to operate on a request body without query/route parameters mixed into it.  It feels awkward to have the `body_parsers` dsl available but only be able to access the result as part of `#params`.

jodosha:
@Moratorius Thanks for this enhancement. May I ask you what's your specific use case for this? I think this PR makes sense, but I'm curious about your need. Thanks. 😄

PR #128, opened 2016-11-11
Avoid symbolizing keys of parsed body.

beauby:
Some middleware may need to access the pristine parsed body (via `env['router.parsed_body']`). Currently, the keys of this hash get symbolized (which is desirable for `env['router.params']` for consistency).

This PR makes it so that the keys are only symbolized in `env['router.params']`, and not in `env['router.parsed_body']`.


jodosha:
Ref #128

jodosha:
@beauby Thanks this looks good and it's the right thing to to. Merging it right away!

PR #134, opened 2016-12-26
:as option for RESTful routes

igneus:
Scenario:

I am developing an application with Czech-only user interface and I want the URLs to be Czech, too. But the code is English, because it's beautiful and convenient. So I define routes like

```ruby
resources :kocky, controller: 'cats'
resources :psi, controller: 'dogs'
```

with Czech resource names mapping to English-named controllers.
The generated route names are not very satisfactory. Not only Czech and English are being mixed together, but furthermore the English pluralizer disfigures the Czech words:

```
             kockies GET, HEAD  /kocky                         Web::Controllers::Cats::Index 
           new_kocky GET, HEAD  /kocky/new                     Web::Controllers::Cats::New   
             kockies POST       /kocky                         Web::Controllers::Cats::Create
               kocky GET, HEAD  /kocky/:id                     Web::Controllers::Cats::Show  
          edit_kocky GET, HEAD  /kocky/:id/edit                Web::Controllers::Cats::Edit  
               kocky PATCH      /kocky/:id                     Web::Controllers::Cats::Update
               kocky DELETE     /kocky/:id                     Web::Controllers::Cats::Destroy
              psuses GET, HEAD  /psi                           Web::Controllers::Dogs::Index 
            new_psus GET, HEAD  /psi/new                       Web::Controllers::Dogs::New   
              psuses POST       /psi                           Web::Controllers::Dogs::Create
                psus GET, HEAD  /psi/:id                       Web::Controllers::Dogs::Show  
           edit_psus GET, HEAD  /psi/:id/edit                  Web::Controllers::Dogs::Edit  
                psus PATCH      /psi/:id                       Web::Controllers::Dogs::Update
                psus DELETE     /psi/:id                       Web::Controllers::Dogs::Destroy
```

I propose adding option `:as` (analogically to the non-RESTful routes) that will allow us to specify an alternative resource name to be used for internal route names instead of the public resource name used in URLs.

```ruby
resources :kocky, controller: 'cats', as: 'cats'
resources :psi, controller: 'dogs', as: 'dogs'
```

### DIFF
--- a/lib/hanami/routing/resource/action.rb
+++ b/lib/hanami/routing/resource/action.rb
@@ -258,7 +258,8 @@ module Hanami
         # @api private
         # @since 0.4.0
         def _singularized_as
-          resource_name.split(NESTED_ROUTES_SEPARATOR).map do |name|
+          name = @options[:as] ? @options[:as].to_s : resource_name
+          name.split(NESTED_ROUTES_SEPARATOR).map do |name|
             Hanami::Utils::String.new(name).singularize
           end
         end
--- a/test/named_routes_test.rb
+++ b/test/named_routes_test.rb
@@ -9,6 +9,8 @@ describe Hanami::Router do
     @router.get('/books/:id',   id: /\d+/, to: endpoint, as: :constraints)
     @router.get('/articles(.:format)',     to: endpoint, as: :optional)
     @router.get('/files/*',                to: endpoint, as: :glob)
+    @router.resources(:leaves,                           as: :resources)
+    @router.resource(:stem,                              as: :singular_resource)
   end
 
   after do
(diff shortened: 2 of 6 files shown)


AlfonsoUceda:
@igneus thanks for this PR, I think you are fixing a bug because we have the `:as` option but it was working for routes using the verb method, e.g:
```ruby
router = Hanami::Router.new
endpoint = ->(env) {
  [200, {}, ["hello"]]
}
router.get('/kocky', to: endpoint, as: :cats).name # => :cats
router.path(:cats) # => "/kocky"
puts router.inspector

# Name Method     Path                           Action
# cats GET, HEAD  /kocky                         #<Proc@/Users/alfonsouceda/code/hanami_official/router/(irb):3 (lambda)>
```

what do you think @jodosha ?

igneus:
@AlfonsoUceda Neither in code, nor in documentation have I found any trace of an intent to support the `:as` option for resource routes. So I definitely wouldn't call this PR a bug fix. It's rather a contribution towards feature symmetry between resource routes and "plain" routes. :)

AlfonsoUceda:
@igneus, you are right, I said it because I was trying using it and nothing happened, so yes it can be considered as a feature.

AlfonsoUceda:
I have to say for me it's ok and it will add value for the router ;)

AlfonsoUceda:
@igneus could you add tests for what @jodosha suggested? thanks!

igneus:
@AlfonsoUceda @jodosha Sure. I'll get to it in a few days, probably on Saturday. Is it ok? Or is there e.g. a planned release date approaching that would require to speed things up a bit?

AlfonsoUceda:
@igneus it's ok, thanks for you time ;)

igneus:
@jodosha I've attempted to cover everything you mentioned, except of [Nested RESTful resources](https://github.com/hanami/router/blob/master/test/nested_resources_test.rb). There I'm not sure which cases should be covered (so that risks are managed and tests aren't unnecessarily bloated). Is it enough to test e.g.

* anything > named resources
* anything > named resource
* named resources > anything
* named resource > anything

?

PR #137, opened 2017-01-07
Make hanami-router Rack 2.0 compatible.

### LINKED ISSUE #136: Make hanami-router compatible with Rack 2.0 only

pascalbetz:
* Update Rack version
* Add a http_router monkey patch to prevent duping the Rack::Request 
* Fix http_router (even though unmaintained, better safe than sorry)

Closes #136


hanami-router.gemspec
```diff
@@ -19,8 +19,8 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 2.3.0'
 
-  spec.add_dependency 'rack',         '~> 1.6'
-  spec.add_dependency 'http_router',  '~> 0.11'
+  spec.add_dependency 'rack',         '~> 2.0.1'
```
jodosha: @pascalbetz This should be just `'~> 2.0'`

lib/hanami-router.rb
```diff
@@ -1 +1,2 @@
 require 'hanami/router'
+require 'hanami/http_router_monkey_patch'
```
jodosha: This is something that concerns `Hanami::Routing::HttpRouter` which is the engine of `Hanami::Router`. This require should be moved in `lib/hanami/routing/http_router.rb`.

pascalbetz:
@jodosha please re-check

PR #139, opened 2017-01-07
Issue 136: Make router Rack 2.0 compatible

### LINKED ISSUE #136: Make hanami-router compatible with Rack 2.0 only

pascalbetz:
* Update Rack version
* Add a http_router monkey patch to prevent duping the Rack::Request
* Fix http_router (even though unmaintained, better safe than sorry)

Closes #136


pascalbetz:
Testing.

PR #141, opened 2017-02-21
Fix issue with deep symbolize for router params

valikos:
Found that in hanami 1.0.0.beta1 construction like `params.get(:data, :attributes)` is not working. 
I found changes in Hanami::Utils for hash modifications. 
Now for deep symbolize need to use `#deep_symbolize!` instead of `#symbolize!`.
Also I fixed a test to check for deep params.


jodosha:
@valikos Thank you, merging this. 👍

Issue #144, opened 2017-03-20
Migrate from Minitest to RSpec


jodosha:
A few rules to follow:

  * If you want to work on this, please leave a comment here. We want to avoid many people to work on the same task.

  * We expect this migration to happen within a week from now.

  * Do not change the tests, only migrate from Minitest to RSpec syntax. If not sure, please ask here.

  * Do not remove the old tests, only add the new ones.

  * Make CI build to use RSpec tests.

  * **Target `develop` branch, instead of `master`.**

  * All the tests artefacts like fixtures, goes under `spec/support`.

  * Try to reflect the structure under `lib/`. Eg. `lib/hanami/view.rb` should be tested under `spec/hanami/view_spec.rb`, while `test/rendering_test.rb` should go under `spec/integration/rendering_spec.rb`.

If you want to take inspiration, please have a look at: https://github.com/hanami/mailer/pull/67

morrme:
It is almost a week from when you posted this, are you still seeking help ?

jodosha:
@morrme Yes, we do.

jodosha:
@morrme Are you working on this?

morrme:
@jodosha i am working on /helpers, just going over the files before i PR

jodosha:
@morrme Thanks for the update 👍

morrme:
OK, I have completed the files for this one now as well. I may need some help with the file structure, but in the meantime I will set up the PR to see if any other changes need to be made.

morrme:
I am researching the remaining problems, and I welcome advice: 

--the test cases (generate and recognition) may need to be totally rewritten
--there's a test that uses the "must_be_same_as" matcher that doesn't have a good synonym in RSpec. 
Example here: https://github.com/hanami/router/blob/master/test/routing_test.rb#L102

I have been experimenting with changing the "response" and using an "include" matcher, as well as creating a variable for the request and then testing for specific values (body, status, etc) , but I didn't want to change too much without approval. 

These problems constitute over 100 failures in the build, so solving them will go a long way towards completing this issue.

jodosha:
@morrme 
> the test cases (generate and recognition) may need to be totally rewritten

Why, specifically?

> there's a test that uses the "must_be_same_as" matcher that doesn't have a good synonym in RSpec.

```ruby
expect(a).to be(b)
```

Shouldn't this work? `be` assets for object identity.

jodosha:
@morrme 
[...]

PR #145, opened 2017-04-05
Migrate from MiniTest to RSpec , issue #144

morrme:


morrme:
@jodosha  Seems to be stuck on 
`/home/travis/build/hanami/router/spec/hanami/routing/endpoint_resolver_spec.rb:1:in `<top (required)>': uninitialized constant Hanami::Routing (NameError)` 

...and I cannot  find the cause of this one.

jodosha:
@morrme You forgot to generate the `.rspec` file which lets you to require `spec/spec_helper.rb` without explicitly require it from `spec/hanami/routing/endpoint_resolver_spec.rb`.

`spec/spec_helper.rb` requires `hanami-router`, which initializes that module.

PR #149, opened 2017-07-05
Ensure Hanami::Router#redirect to be compatible with #recognize

### LINKED ISSUE #147: redirect route destination not wrapped in `Routing::Endpoint`
Hey folks,

@padget stumbled on a bug in hanami-router. After some digging around I found out what's broken. So here's the issue report.

Suppose you have the following routes definition:

```ruby
redirect '/redirect-me', to: '/to-here'
get '/to-here', to: ->(env) { [200, {}, ['Hello from Redirect-Target']] }
```

When you visit `/redirect-me` in the browser you will get redirected. And everything seems to work.
However, when you want to write a test case like this:

```ruby
RSpec.describe Web.routes do
  it 'recognizes "GET /redirect-me"' do
    env   = Rack::MockRequest.env_for('/redirect-me')
    route = described_class.recognize(env)

    expect(route).to be_routable
  end
```

You will receive this error:

```ruby
NoMethodError: undefined method `routable?' for #<Proc:0x007f89c7227628>
```

The reason is that the destination of the recognized route is not wrapped in a `Hanami::Routing::Endpoint` for the `Hanami::Router::redirect` call. It seems that the redirect keyword was not considered when the `Hanami::Routing::Endpoint` classes where introduced. I was unable to find a good way to fix this with minimal changes.

For all the other methods (get, post, root, etc.) the flow is that the corresponding http verb method is called on `Hanami::HttpRouter` which returns a `Hanami::Routing::Route` by going through `Hanami::Routing::Route#generate` which gets a `Hanami::Routing::EndpointResolver` passed whose job it is - among other things - to wrap the `Hanami::Routing::Route` `@dest` in a `Hanami::Routing::Endpoint` depending on the endpoint type.

[Redirect](https://github.com/hanami/router/blob/2c2356b70e73eb4fd3476ecdb6f1d7a3db22bfa3/lib/hanami/router.rb#L625), however, calls `::HttpRouter::Route#redirect` on the result of `Hanami::HttpRouter#get` and that method overrides the `@dest` instance variable of the `Hanami::Routing::Route`. The `Hanami::Routing::EndpointResolver` is not visible in either `Hanami::Routing::Route` or `Hanami::Router`, it is only known to the `Hanami::Routing::HttpRouter` who is no longer involved.

I hope I could explain this in a manner that is easy to follow. It was quite difficult gathering all the moving pieces to locate the issue.

jodosha:
Fix #147

### DIFF
--- a/spec/unit/hanami/router/recognize_spec.rb
+++ b/spec/unit/hanami/router/recognize_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe Hanami::Router do
       end
     end
 
-    describe 'from Rack env' do
+    context 'from Rack env' do
       it 'recognizes proc' do
         env   = Rack::MockRequest.env_for('/proc', method: :get)
         route = @router.recognize(env)
@@ -22,8 +22,10 @@ RSpec.describe Hanami::Router do
 
         expect(body).to eq(['OK'])
 
-        expect(route.routable?).to be true # Expected route to be routable
+        expect(route.routable?).to be(true)
+        expect(route.redirect?).to be(false)
         expect(route.action).to include('spec/unit/hanami/router/recognize_spec.rb:9 (lambda)')
+        expect(route.redirection_path).to be(nil)
         expect(route.verb).to eq('GET')
         expect(route.path).to eq('/proc')
         expect(route.params).to eq({})
@@ -33,8 +35,10 @@ RSpec.describe Hanami::Router do
         env   = Rack::MockRequest.env_for('/resources/1', method: :get)
         route = @router.recognize(env)
 
-        expect(route.routable?).to be true # Expected route to be routable
+        expect(route.routable?).to be(true)
+        expect(route.redirect?).to be(false)
         expect(route.action).to include('spec/unit/hanami/router/recognize_spec.rb:10 (lambda)')
+        expect(route.redirection_path).to be(nil)
         expect(route.verb).to eq('GET')
         expect(route.path).to eq('/resources/1')
         expect(route.params).to eq(id: '1')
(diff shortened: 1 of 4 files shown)


spec/unit/hanami/router/recognize_spec.rb
```diff
@@ -210,6 +227,21 @@
         expect(route.params).to eq({})
       end
 
+      it 'recognizes redirect' do
```
TiteiKo: I'd see something more like
```ruby
expect(route.redirect?).to be true
expect(route.redirection_path).to eql('/')
```

mereghost: I've noted that there seems to be no love for the `expect(route).to be_redirect` notation? Any particular reasons you folks don't like it?

TiteiKo: @mereghost I don't like it because it's not good english... `expect(route).to redirect_to('/')` would be way better.
Unlike being routable, or truthy, you can't "be redirect": a request can "be redirected" or a route can "be a redirection".

jodosha: Oh thanks for the suggestion, I wanted to have `#redirect?` as private method to make `RecognizedRoute#action` to return `nil` in case of redirect. But yeah that can make a great way to inspect the route. Let's add it! 😄

jodosha:
@TiteiKo I've added the new introspection methods, please review again. 😄

kaikuchn:
@jodosha Finally found some time to look it over. It looks good to me!

jodosha:
@kaikuchn Thanks for getting back to us. We've just released a `hanami-router-1.0.1` with that fix and enhancement. Enjoy! 👍

PR #157, opened 2017-12-01
Single tier routing for Hanami

jodosha:
**TL;DR:** In Hanami 1.x series we have inefficient routing, we want to make it efficient in 2.x.

---

## Existing

We call the existing Hanami routing a **two-tier routing**. The first tier is for the Hanami project (let's say a `bookshelf` product), and the second tier is within a single app (`Web`, `Admin`, etc..). Imagine this as an "umbrella routing".

### Project tier

When a request to `/admin/people/23` comes in, the first tier (project router) decides which app should be routed to. Let's see this practically:

```ruby
# config/environment.rb
Hanami.configure do
  mount Admin::Application, at: "/admin"
  mount Web::Application, at: "/"
  # ...
end
```

Behind the code above, there is an `Hanami::Router` instance. The main responsibility of this router is to choose the right app, or to return a `404`.

### Application tier

Following the example above, the project tier (router) invokes `Admin::Application#call`, passing the Rack env. At this point, the router for this app (app tier) is responsible to route the request to the action associated to `/people/23` (a subpath of the original request).

Again, if there is an action, it gets invoked, or it returns a `404`.

### Recap

An Hanami project has `n+1` `Hanami::Router` instances, where `n` is the number of apps (`Web`, `Admin`, etc..), plus the project itself (the root of the umbrella).

---

## Proposal

We can use only one `Hanami::Router` at the project level.

When we mount an application, instead of instantiating its own `Hanami::Router`, we can take the routes defined for that app and add to the project router.

With this technique when a request `/admin/people/23` is coming in, the project router is able to pick directly the corresponding action (or `404`), without delegating to a second tier.

This saves us memory: `1` instance of `Hanami::Router` per project, instead of `n+1`. It also saves runtime CPU/memory because it **doesn't** need to go to the routing system twice (one to pick the app, one to pick the action).

### Technical details

This introduces a new API private method: `Hanami::Router#scope` which accepts a path prefix, a Ruby namespace where to lookup for actions and a block that defines the routes.

**Please notice** that this doesn't change how people will mount their apps in Hanami project, the change is happening under the hood.

There is a bit of code repetition between `Hanami::Routing::Prefix` and `Hanami::Routing::Scope` (introduced by this PR). I'm **not** DRYing it for now.

---

Ref https://github.com/hanami/router/pull/156

### DIFF
--- a/lib/hanami/routing/scope.rb
+++ b/lib/hanami/routing/scope.rb
@@ -1,0 +1,111 @@
+# frozen_string_literal: true
+
+require "delegate"
+require "hanami/utils/path_prefix"
+
+module Hanami
+  module Routing
+    # Prefix for routes.
+    # Implementation of Hanami::Router#prefix
+    #
+    # @since x.x.x
+    # @api private
+    #
+    # @see Hanami::Router#prefix
+    class Scope < SimpleDelegator
+      # @api private
+      # @since x.x.x
+      def initialize(router, prefix, namespace, &blk)
+        @router    = router
+        @namespace = namespace
+        @prefix    = Utils::PathPrefix.new(prefix)
+        __setobj__(@router)
+        instance_eval(&blk)
+      end
+
+      def root(to:, as: :root, **, &blk)
+        super(to: to, as: route_name(as), prefix: @prefix, namespace: @namespace, &blk)
+      end
+
+      # @api private
+      # @since x.x.x
+      def get(path, as: nil, **options, &endpoint)
+        super(@prefix.join(path), options.merge(as: route_name(as), namespace: @namespace), &endpoint)
+      end
+
+      # @api private
+      # @since x.x.x
+      def post(path, as: nil, **options, &endpoint)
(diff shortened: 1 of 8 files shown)


jodosha:
To see it in action, please check this commit: https://github.com/hanami/hanami/commit/c91c8d0199523f3435cd0e7b1d106a40f9a433c2#diff-d626299e9ced9a56c9726d8bd95124e3R68

lib/hanami/routing/scope.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      # @api private
+      # @since x.x.x
+      def prefix(path, &blk)
+        super(@prefix.join(path), namespace: @namespace, &blk)
+      end
+
+      private
+
+      ROUTE_NAME_SEPARATOR = "_"
```
artofhuman: maybe better use `private_constant` to make constant really private? =)

jodosha: @artofhuman agreed 👍

PR #158, opened 2017-12-04
Scope configuration

jodosha:
This is a followup of #157. I tried to split the amount of changes between the two PRs, also because this one introduces a breaking change.

## Background

In order to make Hanami faster, since 2.0, the router will hold immutable action instances to serve all the incoming requests. With 1.x, it used to hold the action class and to create a new instance for each request.

Code sketch:

```ruby
# 1.x
@action.class # => Class
@action.name # => Web::Controllers::Home::Index
@action.new.call(env) # new instance for each request

# 2.x
@action.class # => Web::Controllers::Home::Index
@action.call(env) # always the same instance for all the requests
```

## Problem

Given that `hanami-controller` 2.0 needs an instance of `Hanami::Controller::Configuration` to be instantiated, this PR introduces a new keyword `:controller` for `Hanami::Router#initialize`.

## Proposed solution

On 1.x series it was:

```ruby
module Web
  module Controllers
    module Home
      class Index
        include Web::Action

        def call(params)
        end
      end
    end
  end
end

router = Hanami::Router.new(namespace: Web::Controllers) do
  root to: "home#index"
end
```

On 2.x, I suggest:

```ruby
module Web
  module Controllers
    module Home
      class Index < Hanami::Action
        def call(*)
        end
      end
    end
  end
end

configuration = Hanami::Controller::Configuration.new
router = Hanami::Router.new(configuration: configuration, namespace: Web::Controllers) do
  root to: "home#index"
end
```

Regarding the changes introduced by #157, this PR adds another keyword argument for `Hanami::Router#scope`: `:configuration`.

```ruby
admin = Hanami::Controller::Configuration.new
web = Hanami::Controller::Configuration.new

router = Hanami::Router.new do
  scope "/admin", namespace: Admin::Controllers, configuration: admin do
    # ...
  end

  scope "/", namespace: Web::Controllers, configuration: web do
    # ...
  end
end
```

---

**Please note** that the usage described in this PR it only concerns **standalone usage** of `hanami-router` and `hanami-controller`. Hanami users (`hanami`), won't see the difference.

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -236,16 +236,17 @@ module Hanami
     #   [200, {}, ["{:name=>\"LG\",:id=>\"1\"}"]]
     #
     # rubocop:disable Metrics/MethodLength
-    def initialize(scheme: "http", host: "localhost", port: 80, prefix: "", namespace: nil, parsers: [], force_ssl: false, not_found: NOT_FOUND, not_allowed: NOT_ALLOWED, &blk)
-      @routes      = []
-      @named       = {}
-      @namespace   = namespace
-      @base        = Routing::Uri.build(scheme: scheme, host: host, port: port)
-      @prefix      = Utils::PathPrefix.new(prefix)
-      @parsers     = Routing::Parsers.new(parsers)
-      @force_ssl   = Hanami::Routing::ForceSsl.new(force_ssl, host: host, port: port)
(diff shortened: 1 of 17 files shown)


jodosha:
This PR is a requirement for https://github.com/hanami/controller/pull/245

Issue #160, opened 2018-04-13
Deprecate Force SSL


jodosha:
Force SSL feature makes the internal of `Hanami::Router#call` complicated.

Given there are better Rack based alternatives (as middleware), or NGINX rewriting rules, I don't see the point of keeping this feature here.

  - [x] Deprecate using `Hanami::Utils::Deprecation`

---

This change **MUST** go to `develop` branch.

---

Ref https://trello.com/c/fQw5F4Hp

jodosha:
Implemented by https://github.com/hanami/router/pull/167

PR #167, opened 2018-04-14
Deprecate force ssl use

AlfonsoUceda:
This PR implements #160


spec/integration/hanami/router/force_ssl_spec.rb
```diff
@@ -1,4 +1,14 @@
 RSpec.describe Hanami::Router do
+  before do
+    @original_stderr = $stderr
```
AlfonsoUceda: I think we can move this logic to `Hanami::Utils::IO`, what do you think?

AlfonsoUceda:
@jodosha I did in the specs but I just wanted suppress deprecation warnings when running the whole spec file to avoid noise in the terminal.

PR #169, opened 2018-04-26
remove body cleanup for HEAD requests

### LINKED ISSUE #164: Remove body cleanup for HEAD HTTP requests
`Hanami::Router` 1.x empties the body if the HTTP request is using HEAD verb.

While this is formally correct, and makes the router "full featured", we have the same control in `hanami-controller` and `hanami`. This makes a total of three times for each request. The idea is to use `Rack::Head` from `rack` gem instead of these custom controls.

Developers who are using `hanami-router` as standalone router, can mount that middleware.
Developers who are using `hanami` don't need to take any action, because we'll mount that middleware at the project level.

---

This change **MUST** go to `unstable` branch.

---

https://trello.com/c/fQw5F4Hp

GustavoCaso:
Fixes #164 

@jodosha when removing the code that was cleaning the body for `HEAD` requests, many tests fail because we are using `lint` for `Rack::MockResponse`. 

To fix that I use `Rack::Head` middleware to solve it, also had to add a special case for the `routing_spec` by modifying the body depending on the type of request.

WDYT?

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -1035,10 +1035,7 @@ module Hanami
       if (response = @force_ssl.call(env))
         response
       else
-        enforce_head(
-          env,
-          (@routes.find { |r| r.match?(env) } || fallback(env)).call(@parsers.call(env))
-        )
+        (@routes.find { |r| r.match?(env) } || fallback(env)).call(@parsers.call(env))
       end
     end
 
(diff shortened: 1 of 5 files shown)


jodosha:
@GustavoCaso I merged `unstable` here, because I updated that branch to make the build to pass with the latest `rubocop`. This should make CI to pass for this PR as well.

GustavoCaso:
@jodosha I updated the specs 😉

PR #170, opened 2018-05-01
Path generation bug for regex constraint with leading carrot

cllns:
I just found this bug (Hanami 1.2.0), where specifying a regex with a leading `^` (beginning of string meta-character) will cause the path helpers to fail to find the correct path. 

I have a (currently failing) spec below demonstrating this problem. If you remove the `^`, it works fine. 

(The example also has a trailing `$`, but that's not causing problems.)

In terms of writing this spec, it might make sense to simplify the spec by just replacing the `constraints` example with this one, which I've called `slug_constraints`.

### DIFF
--- a/spec/unit/hanami/router/path_spec.rb
+++ b/spec/unit/hanami/router/path_spec.rb
@@ -9,6 +9,10 @@ RSpec.describe Hanami::Router do
     @router.get('/files/*',              to: endpoint, as: :glob)
     @router.resources(:leaves,                         as: :resources)
     @router.resource(:stem,                            as: :singular_resource)
+    @router.get('/books/:slug',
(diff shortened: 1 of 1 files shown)


cllns:
It looks like this could be a bug in `http_router`, which the current version of `hanami-router` depends on, and isn't maintained :/ 

It looks like the leading carrot isn't needed anyway. Without it, it still treats the beginning of the regex as the beginning of what to match (so 123blah) doesn't work, as we would expect. Going to close this. Hopefully it'll help someone in the future.

jodosha:
@cllns Is it OK to delete the unmerged branch?

cllns:
yes sorry :) just deleted.

PR #175, opened 2018-07-10
Extracted body parser into Rack Middleware

### LINKED ISSUE #162: Extract Body Parsers
Body Parses feature makes the internal of `Hanami::Router#call` complicated.

We can extract into a Rack middleware and invite developers to mount before the request will hit the router.

  - [ ] Extract body parsers into a Rack middleware
  - [ ] Deprecate `:parsers` option for `Hanami::Router#initialize` using `Hanami::Utils::Deprecation`

---

This change **MUST** go to `develop` branch.

---

Ref https://trello.com/c/fQw5F4Hp

GustavoCaso:
Fix #162 

# What 
- [x] Deprecate `options[:parsers]`
- [x] Extract body parser to Rack Middleware
- [x] Add Documentation
- [x] Update README (once we are happy with the implementation I will update it)
- [x] Discuss with @jodosha if we like this interface, so far I have gone this direction because I wanted to read more about internals of Rack itself, but I will like to investigate as well using:
```ruby
use Hanami::Routing::Middleware::BodyParser.new(:json, XmlParser)
```

I went with the block approach because at the end what we are doing is introducing multiple middlewares so `Rack::Builder` really shine here. If we do not like this interface I can change 😄 

I have been looking at the documentation for how to add middleware into hanami:

```ruby
Hanami.configure do
  middleware.use Hanami::Routing::Middleware::BodyParser do
    add_parser :json
  end
end
```
Looks OK to me.


GustavoCaso:
@jodosha I wasn't able to use `devtools` gem for testing the `Deprecation` warning. I added in the `Gemfile` and used `silence_deprecations: true` but it wasn't working, did anything changed?

lib/hanami/routing/http_router.rb
```diff
@@ -67,6 +67,7 @@ class HttpRouter < ::HttpRouter
       # @since 0.1.0
       # @api private
       def initialize(options = {}, &blk)
+        Hanami::Utils::Deprecation.new('options[:parsers] is deprecated and it will be removed in future versions')
```
AlfonsoUceda: I think this message should be displayed only if `:parsers` is used. In addition, we have to say where is the reference of this option, something like this? 

`Hanami::Router options[:parsers]`

lib/hanami/routing/middleware/body_parser.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'hanami/utils/json'
+require 'hanami/routing/middleware/parser'
+
+module Hanami
+  module Routing
+    module Middleware
+      class JsonParser < Parser
```
AlfonsoUceda: better use a separated class?

spec/integration/hanami/router/body_parsing_middelware_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    hanami_app = Hanami::Routing::Middleware::BodyParser.new(@routes) do
+      add_parser :json
+      add_parser XmlMiddelwareParser
+    end
+
+    @app = Rack::MockRequest.new(hanami_app)
+  end
+
+  it 'is successful (JSON)' do
```
AlfonsoUceda: what do you think to create a context for JSON and XML?

spec/integration/hanami/router/middleware/parser_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    let(:env) { Rack::MockRequest.env_for('/', method: 'POST', 'CONTENT_TYPE' => content_type, input: body) }
+    let(:body)         { '' }
+    let(:content_type) { '' }
+
+    describe 'and a JSON request' do
+      let(:body)         { %({"attribute":"ok"}) }
+      let(:content_type) { 'application/json' }
+
+      it "parses params from body" do
+       middleware.call(env)
```
AlfonsoUceda: indentation 👀

spec/integration/hanami/router/middleware/parser_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        it "stores parsed body" do
+           middleware.call(env)
+          expect(env['router.parsed_body']).to eq(["foo"])
+        end
+      end
+
+      describe 'with malformed json' do
+        let(:body) { %({"hanami":"ok" "attribute":"ok"}) }
+        it 'raises an exception' do
```
AlfonsoUceda: space between this line and previous one?

jodosha:
@GustavoCaso I merged `develop` here, now you should be able to use `hanami-devtools` with the deprecation warning silencer.

jodosha:
@GustavoCaso Regarding the DSL for `hanami`, we still need to support the following syntax:

```
# apps/web/application.rb
module Web
  class Application < Hanami::Application
    configure do
      body_parsers :json, XmlParser.new
    end
  end
end
```

We can't introduce breaking changes in 1.3.

That's why I think the syntax without the block it's easier to manage, because internally we need to mount the new middleware.

spec/unit/hanami/routing/http_router_spec.rb
```diff
@@ -26,6 +26,13 @@ def rack_request
     end
   end
 
+  describe '#new' do
+    it 'shows deprecation warning regarding options[:parsers]'do
```
jodosha: There is a space missing here 😉

spec/unit/hanami/routing/error_spec.rb
```diff
@@ -6,6 +6,8 @@
   it 'is parent to all custom exception' do
     expect(Hanami::Routing::Parsing::BodyParsingError.superclass).to eq(Hanami::Routing::Error)
     expect(Hanami::Routing::Parsing::UnknownParserError.superclass).to eq(Hanami::Routing::Error)
+    expect(Hanami::Routing::Middleware::BodyParsingError.superclass).to eq(Hanami::Routing::Error)
```
jodosha: We can't introduce new exceptions.

Imagine someone is doing:

```ruby
begin
  # ...
rescue Hanami::Routing::Parsing::BodyParsingError
end
```

Then suddenly the body parser raises `Hanami::Routing::Middleware::BodyParsingError` instead. The `rescue` block won't work anymore. This is a breaking change.

lib/hanami/routing/middleware/parser.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+require 'hanami/utils/hash'
+
+module Hanami
+  module Routing
```
jodosha: My suggestion is to provide a single Ruby namespace for all the Hanami middleware, despite the gem they are distributed with. So for instance, `Hanami::Middleware::BodyParsers` is distributed by `hanami-router`, and another (fictional) middleware like `Hanami::Middleware::DoSomethingWithAssets` can be distributed by `hanami-assets`.

In this way both `hanami` and final users know for sure that `Hanami::Middleware` hosts all the Rack middleware.

lib/hanami/routing/http_router.rb
GustavoCaso: Thank you @AlfonsoUceda I will change it

spec/integration/hanami/router/body_parsing_middelware_spec.rb
GustavoCaso: Sure no problem

lib/hanami/routing/middleware/parser.rb
GustavoCaso: I really like this idea I will move the code to that namespace

lib/hanami/routing/middleware/body_parser.rb
GustavoCaso: What would be the benefit?

GustavoCaso:
@jodosha @AlfonsoUceda I have applied the feedback.

I have moved away from the block version and use the 
`use Hanami::Middleware::BodyParser.new([:json, XmlParser])` way both for compatibility reasons and for perf reason discussed with @jodosha 😄 

The `BodyParser` itself returns a Rack Middleware, there is some curious thing about how to access the parsers from within the middleware, hope my solution is up for the task. I also added an extra safeguard, if the class passed to the `BodyParser` do not inherit from  `Hanami::Middleware::Parser` it raises an error.

Looking forward to reading your comments.

GustavoCaso:
@jodosha here is a gist https://gist.github.com/GustavoCaso/c6e859819320f4659571fc3f3a925e18 that test the use of the middleware in a standalone in a rack app, thanks to it I found a bug that it was not covered in the specs 😄

PR #176, opened 2018-07-24
remove body parsers

### LINKED ISSUE #163: Remove Body Parsers
Body Parsers were deprecated in v1.3.0

  - [ ] Remove body parsers from `Hanami::Router`

---

This change **MUST** go to `unstable` branch.

---

Ref https://trello.com/c/fQw5F4Hp
Ref https://github.com/hanami/router/issues/162

GustavoCaso:
Fixes #163 

Removes body parsesr from router

Breaking change: **NO LONGER EXISTS**
- `Hanami::Routing::Parsing::UnknownParserError`
- `Hanami::Routing::Parsing::BodyParsingError`

New Errors
- `Hanami::Middleware::BodyParser::UnknownParserError`
- `Hanami::Middleware::BodyParser::BodyParsingError`


lib/hanami/middleware/body_parser/parser.rb
```diff
@@ -11,11 +11,15 @@ class BodyParser
       # This is raised when parser fails to parse the body
       #
       # @since x.x.x
-      class BodyParsingError < Hanami::Routing::Parsing::BodyParsingError
+      class BodyParsingError < Hanami::Routing::Error
```
jodosha: @GustavoCaso This error is `Hanami::Middleware::BodyParsingError`, it shouldn't inherit from an error from another namespace like `Hanami::Routing`.

---

We should probably create a `Hanami::Middleware::Error` in a third party lib like `hanami-utils`. So it will be available to all the gems that need to provide a middleware for Hanami.

GustavoCaso: Hi @jodosha great catch, I will add the `Hanami::Middleware::Error` to `hanami-utils` and update the code to use it 😄

lib/hanami/router.rb
```diff
@@ -1039,7 +1038,7 @@ def mount(app, at:, host: nil)
     #
     # @since 0.1.0
     def call(env)
-      (@routes.find { |r| r.match?(env) } || fallback(env)).call(@parsers.call(env))
+      (@routes.find { |r| r.match?(env) } || fallback(env)).call(env)
```
jodosha: I love how it's simplified now 💯

Issue #179, opened 2018-09-14
BodyParser's fallback "_" key in env["router.params"] unnecessary for typical x-www-form-urlencoded bodies?


timriley:
Hi gang (/cc @jodosha @GustavoCaso)

Today I [pushed a branch to hanami/controller](https://github.com/hanami/controller/compare/unstable...update-specs-for-router-body-parser-middleware) with updated specs for the new `BodyParser` middleware that was introduced here in v1.3.0.beta1.

Making this change raised some new failing specs in hanami/controller. Here's an example:

```
Failures:

  1) Hanami::Router integration resource calls POST create
     Failure/Error: expect(response.body).to   eq(%({:identity=>{:avatar=>{:image=>\"jodosha.png\"}}}))

       expected: "{:identity=>{:avatar=>{:image=>\"jodosha.png\"}}}"
            got: "{:identity=>{:avatar=>{:image=>\"jodosha.png\"}}, :_=>\"identity[avatar][image]=jodosha.png\"}"

       (compared using ==)
     # ./spec/integration/hanami/controller/routing_spec.rb:52:in `block (3 levels) in <top (required)>'
```

As you can see, the `"router.params"` key set on the rack request env is getting a new `"_"` key set on it as part of the way `BodyParser#_symbolize` works:

```ruby
      # @api private
      def _symbolize(body)
        if body.is_a?(Hash)
          Utils::Hash.deep_symbolize(body)
        else
          { FALLBACK_KEY => body }
        end
      end
```

If the body isn't a hash, it can't be symbolised, so it assigns it in full to a `_` key on a new hash.

I presume this is to support e.g. things like JSON arrays that are submitted to an endpoint.

However, this causes some jank to appear for typical x-www-form-urlencoded request bodies. These are all just plain strings, so the body parser treats them in the same way and assigns them to the `_` key.

However, later on, rack itself knows how to parse them and make them accessible as a params hash. But because we've assigned a copy of the body to `_`, we end up with its original string representation sticking around on the params hash. This feels like an unwanted behaviour to me.

Would you agree?

I think a solid fix might actually be to change the way we do the parsing/symbolizing in `BodyParser`. Right now we have a "null parser" kind of arrangement, where `@parsers[media_type(env)]` always returns an object with a `#parse` method, even if there is no explicitly matching parser type. The "null parser" in this case is `BodyParser::Parser` and its `#parse` just returns `body` unmodified.

[...]

PR #180, opened 2018-09-14
Skip attempting to parse unknown types in BodyParser middleware (for reference only, to be broken up into 2 smaller PRs)

timriley:
Hi @jodosha @GustavoCaso –

Well, I couldn't help myself, so I went ahead and just made a quick initial change to demonstrate the approach I outlined in #179. I feel the code probably speaks for itself better than my words :)

If you think this is a good approach I'd be happy to tidy up this PR and get it ready for merge.

I think there's some refactoring that can be done as part of this, too, e.g. I don't think we need a base `BodyParser::Parser` class that the other parsers inherit from. This would mean we can ditch the `BodyParser::Parser` class altogether and just move `.for` to being a static module function or something similar.

I'd also like to change `.require_parser` so that it accepts more than just strings. IMO it should accept parser classes or instances, which would make this feature much more flexible - it'd allow people to provide their _own_ parsers without having to reopen the `Hanami::Middleware::BodyParser` namespace from within their own app and implement them there.

Resovles #179


timriley:
(I also realise I need to target `develop` instead of `unstable` here. I'll fix that if you think this PR is worth continuing)

timriley:
Hi @jodosha, I've rebased this onto `develop` now and tidied things up.

I did a little reorganising while finishing this off:

- Moved `BodyParser::Parser.for` to `BodyParser.for`, since if we're expecting each content-type parser to inherit from `BodyParser::Parser`, I don't think it makes sense for each of them to carry this unnecessary extra API.
- Updated `BodyParser.for` so it supports passing through already-initialized Parser instances, which means that when the middleware is loaded, parsers can be provided either by name (`:json`), by class (`MyParsers::JsonParser`) or also by instance (`MyParsers::JsonParser.new(maybe_this_one: "wants some config")`)
- Added unit tests for `BodyParser.for` and refactored the existing integration tests while also making it clearer that no parsing is attempted for unknown content types
- Moved the error class definitions into their own file so we can require it from wherever we need

Let me know if you're happy with this. I have a few smaller code-specific questions, which I'll leave as inline comments.

lib/hanami/middleware/body_parser.rb
```diff
@@ -1,5 +1,6 @@
-require 'hanami/middleware/body_parser/parser'
 require 'hanami/utils/hash'
+require_relative 'body_parser/class_interface'
+require_relative 'body_parser/parser'
```
timriley: I feel like `require_relative` is nicer for requiring "neighbouring" file like this (i.e. ones that are closely situated to the current file, and within the same overall module or concept). AFAIK `require_relative` is also a little faster than a straight-up require. You happy for me to do it like this?

lib/hanami/middleware/body_parser.rb
```diff
@@ -8,15 +9,15 @@ module Middleware
     class BodyParser
       # @since 1.3.0
       # @api private
-      CONTENT_TYPE       = 'CONTENT_TYPE'.freeze
```
timriley: n.b. these constant assignments stoped being fully aligned so I figured it didn't make sense to try and preserve it

lib/hanami/middleware/body_parser/class_interface.rb
```diff
@@ -0,0 +1,52 @@
+require_relative 'errors'
+require_relative 'parser'
+
+module Hanami
+  module Middleware
+    class BodyParser
+      module ClassInterface
+        # @since x.x.x
```
timriley: Not sure what we're mean to be putting for these `@since` doc tags

lib/hanami/middleware/body_parser/class_interface.rb
```diff
@@ -0,0 +1,52 @@
...
+          elsif parser_instance?(parser)
+            parser
+          else
+            raise UnknownParserError.new(parser)
+          end
+        end
+
+        private
+
+        # @since 1.3.0
```
timriley: And since this method is now moved, should probably change this `@since` doc tag? Since these are tagged as private we're OK to move them around, right?

lib/hanami/middleware/body_parser/class_interface.rb
```diff
@@ -0,0 +1,52 @@
...
+
+        # @api private
+        def parser_class?(parser)
+          parser.is_a?(Class) && parser < Parser
+        end
+
+        # @api private
+        def parser_instance?(parser)
+          parser.class < Parser
+        end
```
timriley: I do have some other thoughts about this. Our `BodyParser::Parser` class is really not doing all that much now.

Another approach could be to get rid of it entirely and just add a check here to ensure parser classes conform to a certain API. Something like this:

```ruby
def for(parser)
  parser = 
    if parser_name?(parser)
      require_parser(parser)
    elsif parser.is_a?(Class)
      parser.new
    else
      parser
    end

  ensure_parser parser
  
  parser
end

private

def ensure_parser(parser)
  unless %i[mime_types parse].all? { |method| parser.respond_to?(method) }
    raise UnknownParserError.new(parser)
  end
end
```

This would allow us to get rid of this now pretty useless base class and just use a closer-to-duck-typing kind of check to ensure we have a parser object we can use. The exception name isn't _perfect_ for this kind of case anymore (since it's more of an _invalid_ parser than an _unknown_ parser, but I think we can't change the exception name so easily due to semver commitments?)

Anyway, I feel like this would be the better approach, tbh. It would make things more flexible and allow us to get rid of some unnecessary code.

Would you be happy for me to make that change?

timriley: What the hey, I'm just going to go ahead and make this change now, since codecov is telling me I don't have test coverage of `Parser`.

timriley:
OK, _now_ I'm happy with this.

I do have a question about the exception classes: since these are actually entirely new (i.e. they're in `Hanami::Middleware::BodyParser` namespace), does this mean we _can_ actually change their names or introduce new ones? People won't see these exceptions raised with old code, since they'll only be raised if code is explicitly changed to use the BodyParser middleware.

lib/hanami/middleware/body_parser/class_interface.rb
GustavoCaso: Once is release it will have the version that it was introduced. Hope, I'm not wrong @jodosha

GustavoCaso:
@timriley love ❤️ all the changes and refactoring you have introduced. I'm totally on board with them

About the question regarding the name of the error classes, yes they are new ones and probably the name can be changed, but they will need to still inherit from `Hanami::Routing::Parsing` to avoid breaking changes for now, I guess once is completely remove from the code we can remove the inheritance

timriley:
@jodosha ready for you again, btw :)

timriley:
Changes here are all covered by #181 and #182 now.

PR #182, opened 2018-10-10
Refactor BodyParser content type parser loading

timriley:
This PR reintroduces the BodyParser refactorings I made in #180. This time, these are refactorings _only_, since the related bug fix is now merged via #181.

These refactorings improve BodyParser's content type parser loading:

- Remove `BodyParser::Parser` base class (was providing little value or shared behavior, and since content type parsers were previously inheriting from it, they were picking up unnecessary extra API, via `.for`, which I've now moved, see below)
- Move `BodyParser::Parser.for` to `BodyParser.for`
- Update `BodyParser.for` so it supports passing through already-initialized Parser instances, which means that when the middleware is loaded, parsers can be provided either by name (`:json`), by class (`MyParsers::JsonParser`) or also by instance (`MyParsers::JsonParser.new(maybe_this_one: "wants some config")`)
- Moved BodyParser-related errors into their own source file so they can be required independently as needed

Along with this, add some extra unit testing around the parser loading behavior provided by `BodyParser.for`. This wasn't explicitly covered by unit tests previously.

@jodosha What do you think?


timriley:
Thanks for merging this refactoring, @jodosha.

I'm in mixed minds about this particular idea of using a middleware stack to handle body parsing:

Firstly, I don't think a class with a `#call` method (i.e. making it produce a "callable object") is in any way intrinsically better than a class without one. The classes we have right now are very _specific_ – they're _body parsers_ and nothing more. This is reinforced by their instance methods, `#mime_types` and `#parse` (which also behave in a sane way, i.e. clear input/output, not mutating internal state nor input data). It's also reinforced by the fact that the BodyParser middleware is responsible for marshalling these objects, querying them for mime type matches, invoking them only on a match, and passing them _only_ the data they need (i.e. the body object only). It's a very focused little system. In some ways, this is positive, because it's very clear how things should work, and when new parsers slot in, they should behave in predictably. In other ways, however, the focus of this system means we lose flexibility.

What you're proposing looks to be an increase in this flexibility, but at the cost of focus, i.e. moving from a _specific_ body parsing system to a more _general purpose_ system of request object manipulation.

In doing this, it will mean every content type parser class will need to do some more work (e.g. manage the content type checking themselves, handle the request object passthrough in the non-matching case) and we'll wind up with more duplication of logic (e.g. every content type parser will need to set the right key on the request env with their parsed data, and perhaps every content type parser will need to be responsible for the deep symbolizing of the body data too?). We could perhaps tidy this up be (re)introducing a base class for request body parsers.

Having a general system here does open up more possibilities for us/application authors to do different things in the future.

If we went down this road, I think we wouldn't want to limit this system by adopting it here under the umbrella of this body parsing (rack) middleware. I think we would do ourselves a favour by broadening its remit, and having it be some sort of general request processing middleware stack. Then we'd just need to consider what the configuration looks like for enabling the JSON body parser again in this new world.

What do you think, @jodosha? I feel like something like this could be worth doing, but given it has a broad scope (as I've tried to describe above), we'd want to take a moment to consider how it fits into the broader world of Hanami.

PR #186, opened 2019-02-12
Allow to inject the finding logic via `Hanami::Router#initialize`

jodosha:
## Enhancement

Ditto.

## Reason

Starting from `hanami` `2.0.0.alpha2` we'll have actions registered into `Hanami::Container`. They are loaded before the app router. We want to reuse the action instances from the container, rather than let the router to lookup them again, and store a different instance of each action.

## Tests

Existing tests are enough to cover this change.


jodosha:
@kaikuchn Good point. What do you suggest? Thanks. 😃

timriley:
Hi @jodosha, I'm really happy to see this come in. I actually did a similar piece of work in https://github.com/hanami/router/commit/9de6db76aab38e38b2883f06cbdc32671a8d547d, but unfortunately it was too low on my list to get it tidied up for you in time.

Anyway, perhaps at least a name idea might come from this work. Why don't we call this dependency an "endpoint resolver?" This is a name you _used_ to have in hanami-router, IIRC, and I think it describe's this object's roll perfectly. Better still, we can just use a simple `#call` interface on it since the object's name already describes its purpose.

(n.b. in my commit I also removed/tidied up a bunch of things – that might be worth keeping on your radar too?)

You can actually see a custom endpoint resolver I've already written to work with Hanami-router: https://github.com/icelab/decaf_sucks/blob/master/lib/snowflakes/web/router/endpoint_resolver.rb. Perhaps you might be able to use this (or base your work off it) when you add the resolution of endpoints from a container?

jodosha:
@kaikuchn @timriley Please have a look at the latest commits. The argument is now `endpoint_resolver`, I introduced a new object `Hanami::Routing::Endpoint::Resolver`, which exposes `#call`.

---

I also removed the old `Hanami::Routing::EndpointResolver`, and opted for the nested name, because in `hanami`, I need to inject a different resolver provided by that gem.

I can reuse the same `Hanami::Routing::Endpoint` namespace instead of polluting `Hanami::Routing`.

lib/hanami/router.rb
```diff
@@ -243,17 +243,17 @@ def self.define(&blk)
...
-      @routes        = []
-      @named         = {}
-      @namespace     = namespace
-      @configuration = configuration
-      @base          = Routing::Uri.build(scheme: scheme, host: host, port: port)
-      @prefix        = Utils::PathPrefix.new(prefix)
-      @finder        = finder
-      @inflector     = inflector
-      @not_found     = not_found
-      @not_allowed   = not_allowed
```
timriley: Are the aligned assignments really worth the diff churn here? (I know they were aligned previously, so just one line being out of whack would be kind of strange, but I ask because I wonder if this is an opportunity to remove the alignment and reduce diff churn in the future). I realise this may be a core hanami "style guide" issue I'm touching on here, and if that's actually the case, your answer would help me ensure I follow suit in future :)

jodosha:
@timriley Here's the `hanami` finder https://github.com/hanami/hanami/pull/989/files#diff-f0881d5068f1f38758f237d3e384cb44

lib/hanami/router.rb
jodosha: I like this style. 😸

PR #187, opened 2019-03-08
Update README.md

lowski:
Custom parser example is incorrect - there is no class `Hanami::Middleware::BodyParser::Parser`.

Custom parser classes don't need to extend any class - they can be POROs.


lowski:
Hey @jodosha no problem I will change it.

Thank you for detailed description and I agree with you that extending internal class is better than using PORO, I just wasn't sure which class should I can extend.

PR #192, opened 2019-11-03
Support for Hanami application router

timriley:
Changes required for the new application router to work (as described in https://github.com/hanami/hanami/pull/1019).


timriley:
I need to update tests for this but I wanted to open a PR first as a way to link these changes together.

PR #205, opened 2020-09-02
Add not_found option

katafrakt:
This adds `not_found` option which was present in 1.x and allowed to specify custom handler for when route was not recognized.

~~It was a bit tempting to choose better name (`default_handler` perhaps?), but I decided to stick with original name for this PR.~~

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -17,12 +17,6 @@ module Hanami
     require "hanami/router/block"
     require "hanami/router/url_helpers"
 
-    # Default response when no route was matched
-    #
-    # @api private
-    # @since 2.0.0
-    NOT_FOUND = ->(_) { [404, { "Content-Length" => "9" }, ["Not Found"]] }.freeze
-
     # URL helpers for other Hanami integrations
     #
     # @api private
@@ -702,6 +696,12 @@ module Hanami
     # @since 2.0.0
     # @api private
     PARAMS = "router.params"
+    
+    # Default response when no route was matched
+    #
+    # @api private
+    # @since 2.0.0
+    NOT_FOUND = ->(*) { [404, { "Content-Length" => "9" }, ["Not Found"]] }.freeze
 
     # @since 2.0.0
     # @api private
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -48,6 +48,7 @@ module Hanami
     #   is deployed
     # @param resolver [#call(path, to)] a resolver for route entpoints
     # @param block_context [Hanami::Router::Block::Context)
+    # @param default_app [#call(env)] default handler when route is not matched
     # @param blk [Proc] the route definitions
     #
     # @since 0.1.0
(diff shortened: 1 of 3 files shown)


lib/hanami/router.rb
```diff
@@ -637,7 +639,7 @@ def not_allowed(env)
     # @since 2.0.0
     # @api private
     def not_found
-      [404, { "Content-Length" => "9" }, ["Not Found"]]
+      ->(_) { [404, { "Content-Length" => "9" }, ["Not Found"]] }
     end
```
katafrakt, line 643: This could be a constant, I think, but haven't seen any other constants defined and left it as method.

lib/hanami/router.rb
```diff
@@ -60,13 +61,14 @@ def self.define(&blk)
     #   Hanami::Router.new do
     #     get "/", to: ->(*) { [200, {}, ["OK"]] }
     #   end
-    def initialize(base_url: DEFAULT_BASE_URL, prefix: DEFAULT_PREFIX, resolver: DEFAULT_RESOLVER, block_context: nil, &blk)
+    def initialize(base_url: DEFAULT_BASE_URL, prefix: DEFAULT_PREFIX, resolver: DEFAULT_RESOLVER, block_context: nil, default_app: not_found, &blk)
```
jodosha, line 64: @katafrakt I'd rename the param to `not_found:`.

jodosha: @katafrakt Please pass the default as `NOT_FOUND` (a lambda that returns the default 404 response), and let the `not_found` method to simply invoke `@not_found`.

lib/hanami/router.rb
```diff
@@ -86,7 +88,7 @@ def call(env)
 
       unless endpoint
         return not_allowed(env) ||
-               not_found
+               @default_app.call(env)
```
jodosha, line 91: @katafrakt Please don't change the level of abstraction: `not_found` represents the "not found" high level concept, whereas `@default_app` is an implementation detail. This line should be reverted.

lib/hanami/router.rb
jodosha: @katafrakt It used to be a constant, but it caused a bug: https://github.com/hanami/router/pull/201

If we wrap the array into a lambda then, it's safe to assign the lambda to a constant, as each time it will be invoked, it will instantiate a new array, so we won't recreate the bug above.

```ruby
NOT_FOUND = ->(_) { [404, { "Content-Length" => "9" }, ["Not Found"]] }.freeze
```

lib/hanami/router.rb
```diff
@@ -637,7 +639,7 @@ def not_allowed(env)
     # @since 2.0.0
     # @api private
     def not_found
```
jodosha, line 641: @katafrakt I believe it would be useful to pass `env` to the method signature and than eventually to `@not_found`.

```ruby
def not_found(env)
  @not_found.call(env)
end
```

jodosha (changes requested):
@katafrakt Thanks for the PR, I left comments regarding adjustments that we can make for consistency sake. Thank you 👍

katafrakt:
Thanks for the comments, @jodosha. I addressed them now. Let me know if I should squash the commits before merge or something.

jodosha (approved):
@katafrakt LGTM 👍 Thank you very much for this.

PR #206, opened 2020-09-06
Allow creation of namespaced router without block

wishdev:
Removes the requirement of a blck being provided during the
creation of namespaced routers

For example:

router = Hanami::Router.new
subrouter = router.namespace 'admin'
subrouter.get '/sample', to: Admin::Sample

### DIFF
--- a/lib/hanami/routing/namespace.rb
+++ b/lib/hanami/routing/namespace.rb
@@ -18,7 +18,7 @@ module Hanami
         @router = router
         @name   = Utils::PathPrefix.new(name)
         __setobj__(@router)
-        instance_eval(&blk)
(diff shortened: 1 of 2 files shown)


jodosha (commented):
Hi and thanks for this PR. What's the use case for this?

In the upcoming version 2.0 we won't allow this kind of syntax for a deliberate choice: an router is frozen after the initialization block is evaluated.

What are we missing here? Any case that you have and we didn't foresee?

wishdev:
No particular issue or reason here - coding style only. If you are removing the style across the board - this obvious is no longer relevant.

Thanks for taking a look

PR #210, opened 2021-01-18
Fixed long path etc router inspect output

graudeejs:
When you have a long URL in router, then router inspection output won't have space separating fields.
In this Pull Requests contains a simple fix.

Let me know if I should change anything

EDIT:
Alternatively need to check each columns max length in order to adjust widths before printing

### DIFF
--- a/lib/hanami/router/inspector.rb
+++ b/lib/hanami/router/inspector.rb
@@ -67,16 +67,22 @@ module Hanami
 
       # @api private
       # @since 2.0.0
+      EXTRA_SEPERATOR = " "
+      private_constant :EXTRA_SEPERATOR
+
+      # @api private
+      # @since 2.0.0
       def inspect_route(route)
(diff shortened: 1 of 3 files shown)


jodosha (changes requested):
@graudeejs Thanks for catching this, given you used a single space as padding, I'd make it explicit by extracting in a private constant and use it explicitly in all the lines that you edited.

Can you please verify if we need padding everywhere? E.g. when we have a very long endpoint (`foooooooooooooo/baaaaaaaaaaar/baaaaaaaaaaaaaz#show`) if they don't overlap with named routes (`as:`) and constraints.

Thanks in advance. 👍

graudeejs:
Done.

I'll create another MR with alternative implementation that prints all columns on same line, let's see how that looks and then perhaps decide with pull request to merge.

jodosha:
Closing in favor of https://github.com/hanami/router/pull/211

PR #211, opened 2021-01-19
Refactored routes inspection output

graudeejs:
As discussed in https://github.com/hanami/router/pull/210 I've created 2nd Pull Request with refactored routes inspector output.

This one always writes output in columns. I tried it on my private project and I prefer this output better than in original Pull Requests.
I suggest you try it on your own project to see how it looks (I'd love to share screenshot of my project, but since it's not public yet, I don't want to disclose any info about it).


jodosha:
@graudeejs Thanks for this new PR, I need to pause this enhancement. I discovered that the inspector implementation doesn't work for `hanami-api`. I need to fix that first, and then get back to this.

jodosha:
@graudeejs I fixed the compat problem with `hanami-api`. Would you please rebase `unstable` here and resolve the git conflicts? Thanks.

graudeejs:
Rebased

jodosha (changes requested):
@graudeejs Thanks and sorry to reject this, but the inspection of a route is delegated to `Hanami::Router::Route` now (see the [source code](https://github.com/hanami/router/blob/unstable/lib/hanami/router/route.rb)).

The implementation of `Hanami::Router::Inspector#call` should stay as it was:

```ruby
@routes.map(&:to_inspect).join(NEW_LINE)
```

I'm sorry for the miscommunication.

graudeejs:
I don't see how you could implement pretty printing routes without altering `Hanami::Router::Inspector#call`.

jodosha:
@graudeejs Simply your enhancement should have a new place: the route. 🙂

graudeejs:
@jodosha If I'm reading you correctly, then you're wrong.

Route has fields that may vary in length (`path`, `to`, `constraints`, `name` etc).
A single route can now know optimal padding for these fields in order to make consistent output (when inspector is outputing all routes. Route A and route B live in isolation, they don't check each other to figure out how to best present output).

For that reason route inspector calculates required padding for each field of all routes, to find max padding.
Once max padding for each field is know you can print pretty output.

That's pretty much what this MR does.

Maybe I'm understanding `Hanami::Router::Inspector` wrong. From my perspective it is used to give user ability to print routes.

Alternatively I can create `Hanami::Router::PrettyInspector`, but I doubt that will be any better.

graudeejs:
Where's what I want to achieve:
router-inspec
Consistent output.

P.S. I blurred out some stuff, don't want to reveal project name at this time.

graudeejs:
@jodosha I'd love to hear from you.
If you feel this MR is implemented in wrong way, then perhaps close it, I can reimplement this in my own code.

graudeejs:
So I've added hack in my project to pretty print routes using router `2.0.0.alpha5`

Had to monkey path `Hanami::Router`
```ruby
# frozen_string_literal: true

require "hanami/router"

class Hanami::Router
  def to_inspect
     with(inspector: @inspector)
     @inspector.call
  end
end
```

This allows me to create my own inspector and inject it via existing router param `inspector` which is assigned to `nil` by default.

It would be cool if inspect was assigned to new instance of `Hanami::Router::Inspector` and actually used in `#to_inspect` instead of crating new instance in `#to_inspect` method.
This would allow to inject different inspectors potentially allowing to create inspector that prints `json` for example. That may be very useful in some cases.

Still would love to hear feedback from @jodosha, as current inspector fails to print good output in cases when fields are long.

jodosha:
@graudeejs Thanks for your help. I'm closing this as stale. Thanks! 🙏

graudeejs:
It's stale only, because you didn't reply :(

But thank you for you time nevertheless

PR #214, opened 2021-05-05
Github Actions: Add Truffle Ruby to the build matrix

szTheory:
I was able to get the test suite passing locally using truffleruby+graalvm-21.1.0. Would be great to see gradual Hanami support for Truffle Ruby with its promising future for high performance Ruby.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
           - "2.4"
           - "2.3"
           - "jruby"
+          - "truffleruby-head"
     steps:
       - uses: actions/checkout@v1
       - name: Install package dependencies
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
           - "2.4"
           - "2.3"
           - "jruby"
-          - "truffleruby-head"
+          - "truffleruby"
     steps:
       - uses: actions/checkout@v1
       - name: Install package dependencies
@@ -44,3 +44,4 @@ jobs:
         run: bundle install --jobs 4 --retry 3
       - name: Run all tests
         run: script/ci
+        continue-on-error: ${{ startsWith(matrix.ruby, 'truffleruby') }}


jodosha:
@szTheory Which version MRI version corresponds the current `truffleruby-head`?

solnic:
Please remember that actions still don't support "allow-failures" so we definitely don't want to add truffle (or even jruby) to the main CI build. We can have something like `ci-experimental` and put other rubies there just to be able to see if things are passing.

szTheory:
@jodosha From [their release page](https://github.com/oracle/truffleruby/releases): "TruffleRuby aims to be fully compatible with the standard implementation of Ruby, MRI, version 2.6.6"

@solnic From the `ci.yml` file here it looks like JRuby is already being targeted, but I'm not familiar with Github Actions. I noticed Jeremy Evan's [Sequel](https://github.com/jeremyevans/sequel/blob/master/.github/workflows/ci.yml) and [Roda](https://github.com/jeremyevans/roda/blob/master/.github/workflows/ci.yml) both have Truffle in the CI pipeline, and Sequel has a `continue-on-error` provision. Not sure if that would serve the purpose of allow failures you mentioned:

```yaml
continue-on-error: ${{ startsWith(matrix.ruby, 'truffleruby') }}
```

After checking the [guide on setting up TruffleRuby for CI](https://github.com/oracle/truffleruby/blob/master/doc/user/standalone-distribution.md) I noticed there are actually two targets:, `truffleruby` and `truffleruby-head`. So maybe `truffleruby` would be more appropriate to avoid the risk of unpredictable CI build failures.

solnic:
@szTheory you're right. I forgot that the added continue-on-error eventually! Let's see how this build will go.

jodosha:
@szTheory I closed this by mistake as per git branch changes (`master` => `main`).

But, after an internal discussion we decided to not proceed with Truffle Ruby for now.

szTheory:
Thanks I understand. With no intention to debate, I'm curious what were the main reasons? And does that mean it would also be off the table for `dry-*`? My main problem with Truffle Ruby is the slow startup times which results in a slow CI. The TR dev team says they are working on it but it's not here yet and could be a while. If I had to guess it would be that.

solnic:
@szTheory I'll re-add TruffleRuby and jRuby builds as a separate workflow that won't be disrupting development in dry/rom. It will serve as a reference so that we can at least see if the builds are passing or not. Officially, both implementations won't be supported until further notice. This will be eventually properly documented in READMEs. Stay tuned for more info.

szTheory:
Great to hear thanks!

PR #216, opened 2021-06-19
Monitoring

jodosha:
## Feature

Introduce monitoring concept in `Hanami::Router`.

Added a new dependency for the router that can be specified via the `monitoring:` keyword argument for `Hanami::Router#initialize`.
The object must respond to `#call(*, &blk)`.

This PR ships with a default monitoring implementation based on `dry-events`.

```ruby
# Gemfile

gem "hanami-router"
gem "dry-events"
```

```ruby
require "hanami/router"
require "hanami/router/monitoring"

router = Hanami::Router.new(monitoring: Hanami::Router::Monitoring.new) do
  root { "Hello" }
end

router.subscribe("hanami.monitoring.router.lookup") do |event|
  puts "Router took #{event.payload.fetch(:elapsed)}ms to lookup"
end

router.call(env)
# => "Router took 0.005ms to lookup"
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,5 +10,3 @@ end
 
 gem "hanami-utils", "~> 2.0.alpha", require: false, git: "https://github.com/hanami/utils.git", branch: "main"
 gem "hanami-devtools",              require: false, git: "https://github.com/hanami/devtools.git", branch: "main"
-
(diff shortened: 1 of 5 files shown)


jodosha:
Stale.

PR #217, opened 2022-02-02
Parse form submission and make it available in Router params

jodosha:


lib/hanami/router.rb
```diff
@@ -868,7 +869,13 @@ def _redirect(to, code)
     def _params(env, params)
       params ||= {}
       env[PARAMS] ||= {}
-      env[PARAMS].merge!(Rack::Utils.parse_nested_query(env["QUERY_STRING"]))
+
+      if (input = env[Rack::RACK_INPUT]) and input.rewind
```
cllns, line 873: Is the `and` here instead of `&&` intentional? I can see how it signifies lower precedence than the assignment in the parens, so just wondering

cllns (approved):
One inline question and also:

Will we later expand the router to do the same for JSON params?

lib/hanami/router.rb
jodosha: @jodosha Yes, it's intentional. To me reads: "if rack input is present, then rewind it, and enter the if statement".

jodosha:
@cllns 
> Will we later expand the router to do the same for JSON params?

We already have `Hanami::Middleware::BodyParser` for JSON parsing of request body.

PR #221, opened 2022-06-15
Extract formatter algorithm from the inspector

waiting-for-dev:
Up to this point, the route inspector always returned a visually
attractive representation of the routes. Now, that format has been
extracted to a `Hanami::Formatter::HumanFriendly` class, while others
are allowed to be injected.

That allows the creation of other algorithms that can be helpful for
purposes beyond a visual inspection, like CSV, JSON, or import formats
for API documentation and platforms.

As others can create their formatters, we're changing the visibility of
`Hanami::Inspector` and `Hanami::Route` methods to be public.

We've taken the occasion to refactor the extracted code, so it doesn't
mutate a variable. We've also removed the test on
`Hanami::Inspector#initialize` as it stated something obvious, and the
description was out of date since https://github.com/hanami/router/commit/9936225e36814fbed215d491479597c62b397b66 (no longer
frozen).

There's a second commit fixing a Rubocop offense to make CI green.

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -27,7 +27,6 @@ module Hanami
 
     # Routes for inspection
     #
-    # @api private
     # @since 2.0.0
     attr_reader :routes
 
@@ -603,7 +602,7 @@ module Hanami
       )
     end
 
-    # Returns formatted routes
+    # Returns formatted routes with the default formatter
     #
     # @return [String] formatted routes
     #
--- a/lib/hanami/router/route.rb
+++ b/lib/hanami/router/route.rb
@@ -5,24 +5,27 @@ require "hanami/router/block"
 
 module Hanami
   class Router
+    # A route from the router
+    #
+    # @since 2.0.0
     class Route
       # @api private
       # @since 2.0.0
+      ROUTE_CONSTRAINT_SEPARATOR = ", "
+      private_constant :ROUTE_CONSTRAINT_SEPARATOR
+
+      # @since 2.0.0
       attr_reader :http_method
 
-      # @api private
       # @since 2.0.0
       attr_reader :path
 
-      # @api private
       # @since 2.0.0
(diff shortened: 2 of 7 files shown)


lib/hanami/router.rb
```diff
@@ -27,7 +27,6 @@ class Router
 
     # Routes for inspection
     #
-    # @api private
```
jodosha, line 30: @waiting-for-dev Why?

lib/hanami/router/route.rb
```diff
@@ -5,24 +5,27 @@
 
 module Hanami
   class Router
+    # A route from the router
+    #
+    # @since 2.0.0
     class Route
       # @api private
+      # @since 2.0.0
+      ROUTE_CONSTRAINT_SEPARATOR = ", "
```
jodosha, line 14: @waiting-for-dev Leftover?

jodosha (approved):

timriley (approved):

lib/hanami/router.rb
waiting-for-dev: If you want to use a formatter different than the default one, you need to do something like:

```ruby
# `router` is an instance of `Hanami::Router`
Hanami::Router::Inspector.new(router: router.routes, formatter: my_formatter).call
```

Does it make sense?

lib/hanami/router/route.rb
waiting-for-dev: Actually, it's still [used in the same file](https://github.com/hanami/router/blob/4caec5d79f2449d8bf274d2bac319930522423d7/lib/hanami/router/route.rb#L76) to inspect the constraints.

waiting-for-dev: In fact, it was a leftover in `Hanami::Router::Formatter::HumanFriendly`. I just removed it from there.

PR #230, opened 2022-10-13
Support configurable body parsers

solnic:
- Allow passing mime types to the constructor of a Parser class
- Change `Parser#mime_types` to be `Parser.mime_types`
- Allow passing parser id along with additional mime types as an argument in the `BodyParser` constructor

### DIFF
--- a/lib/hanami/middleware/body_parser/class_interface.rb
+++ b/lib/hanami/middleware/body_parser/class_interface.rb
@@ -21,7 +21,7 @@ module Hanami
           parser =
             case parser
             when String, Symbol
-              build(require_parser(parser), **config)
+              build(parser_class(parser), **config)
             when Class
               parser.new(**config)
             else
@@ -64,18 +64,22 @@ module Hanami
 
         # @api private
         # @since 1.3.0
-        def require_parser(parser)
-          require "hanami/middleware/body_parser/#{parser}_parser"
+        def parser_class(parser_name)
+          require "hanami/middleware/body_parser/#{parser_name}_parser"
 
-          load_parser!("#{classify(parser)}Parser")
+          load_parser!("#{classify(parser_name)}Parser")
         rescue LoadError, NameError
-          raise UnknownParserError.new(parser)
+          raise UnknownParserError, parser_name
         end
 
+        # @api private
+        # @since 1.3.0
         def classify(parser)
           parser.to_s.split(/_/).map(&:capitalize).join
         end
 
+        # @api private
+        # @since 1.3.0
         def load_parser!(class_name)
           Hanami::Middleware::BodyParser.const_get(class_name, false)
         end
--- a/lib/hanami/middleware/body_parser/class_interface.rb
+++ b/lib/hanami/middleware/body_parser/class_interface.rb
@@ -59,7 +59,9 @@ module Hanami
         # @api private
         # @since 1.3.0
         def ensure_parser(parser)
-          raise InvalidParserError.new(parser) unless PARSER_METHODS.all? { |method| parser.respond_to?(method) }
+          unless PARSER_METHODS.all? { |method| parser.respond_to?(method) }
+            raise InvalidParserError.new(parser)
+          end
         end
 
         # @api private
--- a/lib/hanami/middleware/body_parser/class_interface.rb
+++ b/lib/hanami/middleware/body_parser/class_interface.rb
@@ -100,11 +100,11 @@ module Hanami
 
           begin
             require "hanami/middleware/body_parser/#{parser_name}_parser"
-          rescue LoadError;end
+          rescue LoadError; end
 
           begin
             parser = load_parser!("#{classify(parser_name)}Parser")
-          rescue NameError;end
+          rescue NameError; end
 
           parser
         ensure
--- a/lib/hanami/middleware/body_parser/class_interface.rb
+++ b/lib/hanami/middleware/body_parser/class_interface.rb
@@ -10,12 +10,18 @@ module Hanami
       # @since 1.3.0
       module ClassInterface
         # @api private
+        # @since 2.0.0
+        def new(app, parser_specs)
+          super(app, build_parsers(parser_specs))
+        end
+
+        # @api private
         # @since 1.3.0
-        def for(parser, **config)
+        def build(parser, **config)
           parser =
             case parser
             when String, Symbol
-              self.for(require_parser(parser), **config)
+              build(require_parser(parser), **config)
             when Class
               parser.new(**config)
             else
@@ -25,6 +31,23 @@ module Hanami
           ensure_parser parser
 
           parser
+        end
+
+        # @api private
+        # @since 2.0.0
+        def build_parsers(parser_specs)
+          parsers = Array(parser_specs).flatten(0)
+
+          return {} if parsers.empty?
+
+          parsers.each_with_object({}) do |spec, memo|
+            name, *mime_types = Array(*spec).flatten(0)
+            parser = build(name, mime_types: mime_types)
+
+            parser.mime_types.each do |mime|
+              memo[mime] = parser
+            end
+          end
         end
 
         private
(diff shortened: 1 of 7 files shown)


solnic:
[Support activating built-in body parser middlewares for custom mime types](https://trello.com/c/6ousYBxH/247-support-activating-built-in-body-parser-middlewares-for-custom-mime-types)

spec/integration/hanami/middleware/body_parser_spec.rb
```diff
@@ -6,8 +6,22 @@
...
-    it "raises error" do
+  describe ".new" do
+    it "accepts a parser identificator" do
+      body_parser = Hanami::Middleware::BodyParser.new(app, :json)
+
+      expect(body_parser.instance_variable_get("@parsers")["application/json"])
+        .to be_instance_of(Hanami::Middleware::BodyParser::JsonParser)
+    end
+
+    it "accepts a parser identificator with additional mime-type(s)" do
```
timriley, line 17:
```suggestion
    it "accepts a parser name with additional mime-type(s)" do
```

spec/integration/hanami/middleware/body_parser_spec.rb
```diff
@@ -6,8 +6,22 @@
...
+      body_parser = Hanami::Middleware::BodyParser.new(app, [json: "application/json+scim"])
```
timriley, line 18: Do we actually want two specs here?

1. Like this one, which shows the `json` parser being used with exactly 1 additional mime type (not wrapped in an array), and
2. Another one that shows the `json` parser being used with >1 mime type (which requires them to be wrapped in an array)

Otherwise we have nothing exercising the Array() wrapping code up in `#build_parsers`.

(I know this might sound picky, but I genuinely think it'd be helpful, since I was hoping the specs might show me how the `parser_specs` value was meant to work!)

timriley (commented):
Nicely done, @solnic! This cleanly solves the issue I had before, and hopefully it means that anyone else dealing with funky mime types doesn't get caught out wondering if they have to eject entirely from Hanami's built-in tools.

I've left a range of comments here, hopefully they're not too onerous to address. I do think they'd help get this into the best possible shape before we merge, so please let me know what you think 😄

lib/hanami/middleware/body_parser/class_interface.rb
```diff
@@ -11,13 +11,13 @@ class BodyParser
       module ClassInterface
         # @api private
         # @since 1.3.0
-        def for(parser)
+        def for(parser, **config)
           parser =
             case parser
             when String, Symbol
-              require_parser(parser)
+              self.for(require_parser(parser), **config)
```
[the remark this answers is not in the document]
solnic: @timriley I laughed at this gif so loud that you may have heard it in AU

lib/hanami/middleware/body_parser/class_interface.rb
```diff
@@ -44,7 +44,7 @@ def ensure_parser(parser)
         def require_parser(parser)
```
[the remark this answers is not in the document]
solnic: @timriley I very consciously stayed away from making any additional tweaks/improvements but I am totally open for incorporating them in this PR if you're OK with this too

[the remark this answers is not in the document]
timriley: @solnic Yeah, I reckon let's do it. This is probably the last time in a while this area of the code will get focused attention from us, so let's use this opportunity to make sure we leave it feeling cohesive and easier to understand the next time we visit it :)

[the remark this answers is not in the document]
timriley: Appreciate you keeping it tight in the first place! Certainly helped me understand the changes 👍🏼

spec/integration/hanami/middleware/body_parser_spec.rb
solnic: @timriley great catch, this was actually broken

lib/hanami/middleware/body_parser/class_interface.rb
```diff
@@ -9,15 +9,21 @@ class BodyParser
       # @api private
       # @since 1.3.0
       module ClassInterface
+        # @api private
+        # @since 2.0.0
+        def new(app, parser_specs)
```
timriley, line 14: Would you mind leaving some doc comments here about the expected structures for the parser_specs?

timriley (commented):
This is looking wonderful, @solnic! Thanks for coming back to it, and I'm really glad to hear that I helped discover a bug for you to fix 😊

I think there's just one outstanding thing from my list to be addressed — could we document just what `parser_specs` is meant to look like above the method that first receives them? That name is not self-descriptive enough to help someone who is visiting this code, and I reckon having a couple of examples of this param would help us greatly with ongoing maintenance :)

lib/hanami/middleware/body_parser/class_interface.rb
solnic: @timriley done

lib/hanami/middleware/body_parser/class_interface.rb
```diff
@@ -9,15 +9,38 @@ class BodyParser
...
+        # @example
+        #   Hanami::Middleware::BodyParser.new(->(env) { [200, {}, "app"] }, :json)
+        #
+        #   Hanami::Middleware::BodyParser.new(
+        #     ->(env) { [200, {}, "app"] }, [json: "application/json+scim"]
+        #   )
+        #
+        #   Hanami::Middleware::BodyParser.new(
+        #     ->(env) { [200, {}, "app"] }, [json: ["application/json+scim", "application/ld+json"]]
+        #   )
```
timriley, line 23: This example helps so much, thank you @solnic!!

Just so I can understand, is this what the `parser_specs` arg would look like if we wanted to configure multiple parsers with custom mime types?

```ruby
[
  {json: ["application/json+scim", "application/ld+json"]},
  {xml: ["application/xml+foo", "application/xml+bar"]}
]
```

And am I right in assuming that this is what we'd do if we wanted multiple parsers without any custom mime types?

```ruby
[:json, :xml]
```

timriley: (I actually wonder if the case of multiple parsers with multiple custom mime types would be good to include in the tests)

timriley (approved):
Thanks for including that example, @solnic! It's led me to a couple more questions. If you'd like to make any final tweaks based on those questions, I'd be more than happy for you to do so and then merge whenever you're ready :)

Thanks again for the collaboration here! 🙇🏻‍♂️

lib/hanami/middleware/body_parser/class_interface.rb
timriley: Actually, does all our Array() wrapping allow the parser_specs to be passed like this?

```ruby
Hanami::Middleware::BodyParser.new(env, json: ["custom", "type"], xml: ["custom", "types"])
```

Because that would be ideal. It feels the most direct and natural way to express it.

solnic: @timriley the intention was to support what's passed from the config, I assumed nobody will be initializing a body parser manually, it's an implementation detail and the whole class is actually private. It doesn't make sense to focus on its ergonomics because of this. I would change many things here but it's just not worth the effort.

> Actually, does all our Array() wrapping allow the parser_specs to be passed like this?
> 
> Hanami::Middleware::BodyParser.new(env, json: ["custom", "type"], xml: ["custom", "types"])

I've made this possible:

```ruby
Hanami::Middleware::BodyParser.new(env, [json: ["custom", "type"], xml: ["custom", "types"]])
```

Which would be passed in if you did this in the config:

```ruby
config.middleware.use(:body_parser, [json: ["custom", "type"], xml: ["custom", "types"]])
```

FYI the reason why this is needed to be passed in as an array is because `use` is a method defined in `Middleware::Stack` which supports its own options. This kind of sucks and it was mentioned when I introduced these options (`before` and `after`) but I'm not sure if it makes sense to do anything about it now.

Issue #228, opened 2022-10-10
Surprising prefix behavior


Jammjammjamm:
I use hanami router in a non-hanami application, and I found the current prefix handling to be very confusing. I decided to test its behavior using the following simple `config.ru` file with puma.

```ruby
require 'hanami/router'

router = Hanami::Router.new(prefix: 'abc') do
  get '/', to: ->(env) { [200, {}, ['Success!']]}
end

run router
```
Trying out various prefixes, I see the following results:

|prefix|url|response status|
|---|---|---|
|`abc`|http://localhost:4567/abc|404|
|`abc`|http://localhost:4567/abc/|404|
|`/abc`|http://localhost:4567/abc|200|
|`/abc`|http://localhost:4567/abc/|404|
|`/abc/`|http://localhost:4567/abc|404|
|`/abc/`|http://localhost:4567/abc/|200|

I wouldn't expect to have to use a leading slash in order for the prefix to work. I also don't typically handle requests differently based on whether they contain a trailing slash, but I don't see how to make the router behave that way.

Jammjammjamm:
We're having to work around the trailing slash issue by not using `prefix` at all:
```ruby
route_block = proc do
  scope 'api' ...
  ...
  get '/', to: root_path_handler
end
    
Router =
  if base_path.present?
    Hanami::Router.new do
      scope("#{base_path}/") do
        get '/', to: root_path_handler
      end
      scope(base_path, &route_block)
    end
  else
    Hanami::Router.new(&route_block)
  end

```

jodosha:
@Jammjammjamm Thanks for reporting.
Please remember that `prefix:` must be in form of `"/abc"`, not `"abc"`.

I added specs to ensure this would work as expected: https://github.com/hanami/router/pull/235

PR #231, opened 2022-10-18
Allow router prefix to work with or without a leading slash

chris-at-work:
Addresses #228.  The test I have added is not complete so I'm not sure if you want to duplicate all tests or really reorganize the whole thing when maybe there is not a risk of regressions?


jodosha:
Closing in favor of https://github.com/hanami/router/pull/235

PR #238, opened 2022-11-13
Convert to Zeitwerk for code loading

timriley:
Covert the gem to use Zeitwerk for code loading. To do this:

- Add zeitwerk to the gemspec
- Add `lib/hanami-router.rb` so the gem follows a conventional structure
- Set up the Zeitwerk loader inside `lib/hanami/router.rb` (call it `.gem_loader` instead of `.loader` like we have done in other gems, since `Hanami::Router` is a class with other responsibilities, and `loader` is too generic a term)
- Remove all now-unneeded manual requires for files inside the gem

This is not a breaking change because any existing `require "hanami/router"` will continue to work as expected, since that is where the Zeitwerk loader setup is done.

### DIFF
--- a/hanami-router.gemspec
+++ b/hanami-router.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "rack",               "~> 2.0"
   spec.add_dependency "mustermann",         "~> 1.0"
   spec.add_dependency "mustermann-contrib", "~> 1.0"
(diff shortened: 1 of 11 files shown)


timriley:
[Zeitwerkify hanami-router](https://trello.com/c/C1xQlw6f/329-zeitwerkify-hanami-router)

timriley:
We won't be proceeding with this one, in order to minimise router dependencies given router is used by hanami-api, which is intended to be as minimal as possible.

Issue #237, opened 2022-11-10
Exception when JSON body contains %


Jammjammjamm:
One of our users [reported an error](https://github.com/onc-healthit/onc-certification-g10-test-kit/issues/338) which occurred when they were submitting an AJAX-based form. After some investigation, we discovered that the error was caused by the `%` character in the JSON body of the POST.

Here's a basic spec to demonstrate, it occurs with or without the json body parser:
```ruby
require 'hanami/middleware/body_parser'
require 'hanami/router'
require 'rack/test'
require 'json'

RSpec.describe 'routing wildcard behavior' do
  include Rack::Test::Methods

  let(:app) do
    router =
      Hanami::Router.new do
        post '/abc', to: ->(env) { [200, {}, ['resources/:id']]}
      end
    Hanami::Middleware::BodyParser.new(router, :json)
  end

  it 'responds to /abc' do
    post '/abc', JSON.generate('foo' => '%'), 'CONTENT_TYPE' => 'application/json'
    expect(last_response.status).to eq(200)
  end
end
```
The result:
```
ᐅ be rspec router_spec.rb
F

Failures:

  1) routing wildcard behavior responds to /abc
     Failure/Error: post '/abc', JSON.generate('foo' => '%'), 'CONTENT_TYPE' => 'application/json'

     Rack::QueryParser::InvalidParameterError:
       invalid %-encoding ({"foo":"%"})
     # ./router_spec.rb:18:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # ArgumentError:
     #   invalid %-encoding ({"foo":"%"})
     #   ./router_spec.rb:18:in `block (2 levels) in <top (required)>'

Finished in 0.01218 seconds (files took 0.09398 seconds to load)
1 example, 1 failure
```
Stack trace:
```
~/.asdf/installs/ruby/3.1.2/lib/ruby/3.1.0/uri/common.rb:341:in `decode_www_form_component': invalid %-encoding ({"foo":"%"}) (Rack::QueryParser::InvalidParameterError)
	from ~/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/rack-2.2.4/lib/rack/utils.rb:57:in `unescape'
	from ~/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/rack-2.2.4/lib/rack/query_parser.rb:159:in `unescape'
[...]

PR #239, opened 2022-11-15
Fix parsing of multipart/form-data

wuarmin:
I tried to fix issue #237. This also fixes the issue, posted at [hanami-discourse](https://discourse.hanamirb.org/t/when-i-upload-a-xml-file-i-get-rack-invalidparametererror/719).

I also tried to test a file upload. So far unfortunately without success. It fails with an `EOFError`. Do build the `multipart_fixture` I used some code from `rack`-specs.

WDYT?

Here's my approach:
```ruby
  context "file upload" do
    def multipart_fixture(name, boundary = "AaB03x")
      file = fixture_path(name)
      data = File.open(file, 'rb') { |io| io.read }
  
      type = %(multipart/form-data; boundary=#{boundary})
      length = data.size
  
      { "CONTENT_TYPE" => type,
        "CONTENT_LENGTH" => length.to_s,
        :input => StringIO.new(data) }
    end

    fit "should handle a file upload" do
      env = Rack::MockRequest.env_for(
        "/submit",
        multipart_fixture("foo.xml").merge(method: "POST"),
      )

      subject.call(env)

      puts env["router.params"]
    end
  end
```

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -919,9 +919,14 @@ module Hanami
       params ||= {}
       env[PARAMS] ||= {}
 
-      if (input = env[::Rack::RACK_INPUT]) and input.rewind
-        env[PARAMS].merge!(::Rack::Utils.parse_nested_query(input.read))
-        input.rewind
+      if (input = env[::Rack::RACK_INPUT])
+        case env["CONTENT_TYPE"]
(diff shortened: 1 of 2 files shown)


jodosha:
@wuarmin Thanks for this PR.
Closing in favor of https://github.com/hanami/router/pull/240

PR #242, opened 2022-11-18
Reproduce the problem

jodosha:
WIP 👉 https://github.com/hanami/router/pull/240#issuecomment-1318868644


wuarmin:
That looks good :+1:

lib/hanami/middleware/body_parser/form_parser.rb
```diff
@@ -0,0 +1,43 @@
...
+module Hanami
+  module Middleware
+    class BodyParser
+      # @since 2.0.0
+      # @api private
+      class FormParser < Parser
+        # @since 2.0.0
+        # @api private
+        MIME_TYPES = [
+          "application/x-www-form-urlencoded",
```
wuarmin, line 15: @jodosha I think we don't need this parser for the mimetype `application/x-www-form-urlencoded`. `multipart/form-data` should be enough in my opinion. 

Otherwise we may get the following `params`-results, when `:form`-parser is used, because  `Rack::Multipart.parse_multipart(env)` returns `nil` for a ``application/x-www-form-urlencoded` payload.

```console
{
  :framework => "hanami",
  :is        => "great",
  :_         => nil
}
```
 Apart from that the branch works like a charm. Now files are uploaded correctly. Why the test still fails is another story that we will clear up.

wuarmin (changes requested):

spec/integration/hanami/router/params_spec.rb
```diff
@@ -71,6 +74,28 @@
...
+    subject do
+      r = router
+
+      Rack::Builder.new do
+        use Hanami::Middleware::BodyParser, :form
+        run r
+      end
+    end
+
+    it "handles file upload" do
```
wuarmin, line 87: @jodosha I tried to fix the spec and tried to use `Rack::Multipart.build_multipart` instead of `multipart_fixture`. It seems to work:
```ruby
    it "handles file upload" do
      file = Rack::Multipart::UploadedFile.new(fixture_path("foo.xml"))
      data = Rack::Multipart.build_multipart("file" => file)
      env = Rack::MockRequest.env_for(
        "/submit",
        "CONTENT_TYPE" => "multipart/form-data; boundary=AaB03x",
        "CONTENT_LENGTH" => data.length.to_s,
        method: "POST",
        :input => StringIO.new(data)
      )

      subject.call(env)

      expect(env["router.params"][:file][:filename]).to eq("foo.xml")
      expect(env["router.params"][:file][:tempfile].read).to eq(File.new(fixture_path("foo.xml")).read)
    end
```

wuarmin (changes requested):

lib/hanami/middleware/body_parser/form_parser.rb
jodosha: @wuarmin This is taken from Rack codebase.

spec/integration/hanami/router/params_spec.rb
jodosha: @wuarmin Thank you 👍

PR #248, opened 2023-01-28
Fix find dynamic path segment for midlewares

wafcio:
Issue reported here: https://github.com/hanami/api/issues/29

Explanation:
path segments are saved as String, even dynamic segments started with ":"

### DIFF
--- a/spec/unit/hanami/middleware/node_spec.rb
+++ b/spec/unit/hanami/middleware/node_spec.rb
@@ -27,11 +27,23 @@ RSpec.describe Hanami::Middleware::Node do
 
   describe "#get" do
     context "when segment is found" do
-      it "returns the node" do
-        segment = "foo"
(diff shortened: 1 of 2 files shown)


spec/unit/hanami/middleware/node_spec.rb
```diff
@@ -27,11 +27,23 @@
...
-        segment = "foo"
-        subject.put(segment)
+      context "and segment is defined as symbol" do
+        it "returns the node" do
+          segment = "foo"
+          dynamic_segment = ":bar"
+          subject.put(dynamic_segment)
+
+          expect(subject.get(segment)).to be_kind_of(described_class)
+        end
```
wafcio, line 37: I am not sure about this test, and test from line 41 (old line 30). I am not sure, because when segment doesn't exists It returns also object of Node class (self)

Issue #261, opened 2023-11-21
Leading slashes not kept for mounted Rack apps


pat:
Hi folks

I need to debug this properly, but I'm a bit short on time this week. Figure it's worth flagging nonetheless in case it's something worth investigating prior to the 2.1.0 release.

I have an app with several slices - one of which is mounted at the root, and within that slice there's a Roda app mounted at the root as well (eventually to be migrated to proper Hanami routes/actions, but it's a big migration, so I'd like to take a piecemeal approach).

The leading slash in PATH_INFO is not being passed through to the Roda app - e.g. a URL of `example.com/foo/bar` comes through as `foo/bar` rather than `/foo/bar`. This makes Roda's routing fail.

I've put in place a middleware (to the Roda app) that adds the slash back in, but I figure this could be an issue others face as well? (I'm also debugging another slice that's not being routed to either - unsure if it's the same issue thus far) - I would expect a mounted app to _just work_ - and if Roda expects that slash, perhaps other Rack libraries/apps do as well.

MRI 3.2.2, Hanami 2.1.0.rc2, macOS Sonoma

timriley:
Fixed by #263.

timriley:
Fixed by #263.

PR #263, opened 2023-11-23
fix PATH_INFO/SCRIPT_NAME for root mounts

pat:
If a mounted Rack app is at the root of the Hanami app, it should keep the leading slash in PATH_INFO, and the SCRIPT_INFO value should be blank. As per [the Rack spec on SCRIPT_INFO](https://github.com/rack/rack/blob/main/SPEC.rdoc):

> This may be an empty string, if the application corresponds to the
> “root” of the server.

The existing behaviour for apps mounted in sub-paths was fine - it's just the root path situation that wasn't being handled correctly. (As initially reported in #261)

I don't feel this is a particularly elegant fix (especially the spec changes), so please suggest changes (or, edit my commit!) if so desired.

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -637,11 +637,13 @@ module Hanami
       @mounted.each do |prefix, app|
         next unless (match = prefix.peek_match(env[::Rack::PATH_INFO]))
 
-        # TODO: ensure compatibility with existing env[::Rack::SCRIPT_NAME]
-        # TODO: cleanup this code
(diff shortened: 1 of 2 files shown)


jodosha:
Hey @pat thanks for this fix.

Does it solve #231 for you?

pat:
Ah, I mistagged - I meant #261, sorry! (Which yes, it does fix)

timriley:
Thanks for this fix, @pat, it looks good to me! I'm going to merge it so we can get this into the impending release.

PR #264, opened 2023-11-23
CI: rely on setup-ruby to install

olleolleolle:
...and use a newer actions/checkout (to avoid any GitHub Actions warning about old stuff used)


jodosha (approved):
@olleolleolle Thanks for this contribution. 🙏

PR #265, opened 2023-11-23
CI: Tell dependabot to update GH Actions, weekly

olleolleolle:
This PR in order to avoid stale GitHub Actions versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


katafrakt:
Thanks for the PR and sorry it took so long to get a reply for it.

I believe for changes like this, we would nowadays like to have them synchronized across all the Hanami (and Dry) gems via our [repo-sync](https://github.com/hanakai-rb/repo-sync) (it was not a thing at the time of creating this PR). Could you start a discussion there about this? I'm not sure how Dependabot fits into our general setup. And with recent pinning Github actions to a specific commit ref, I fear that this warning would trigger way too often. So let's explore and discuss this first.

Issue #262, opened 2023-11-23
globbed paths are always evaluated after mounted paths


pat:
Continuing on from #261, I've determined why my slice isn't being picked up - because it's a globbed path, and I can see in Hanami::Router#lookup that mounted paths are given priority. So, in the following case:

```ruby
# in config/routes.rb
slice :foo, at: "/foo"
slice :main, at: "/"
```

```ruby
# in slices/foo/config/routes.rb
get "/*path", to: "foo_stuff"
```

```ruby
# in slices/main/config/routes.rb
mount MiniApp, at: "/"
```

`MiniApp` will always return a response - and so, `/foo/*path` will never be given a chance to respond, even though it is namespaced, and is listed ahead of the main slice.

If a mounted path returns a 404 (or 405), should that be considered a final response, or should it be skipped in case there's a globbed path? Or, could we adjust the logic of `lookup` so it respects namespacing and/or route-ordering of globbed paths?

pat:
Wondering if an alternative order of considerations for lookup should be something like:

1. fixed paths
2. paths with variables
3. non-root mounted paths
4. non-root glob paths (i.e. don't start with "/*")
5. root mounted paths
6. root glob paths

The ordering of 3 & 4 and 5 & 6 are interchangeable, and wouldn't stop people from overlapping root routes, but at least would avoid root globs/mounts getting in the way of their non-root equivalents.

Or... could mounts and globs be grouped together (mounting something at `/foo` is kinda equivalent to a glob of `/foo/*path` after all)? Then all that's needed is for root globs/mounts to be listed after the non-root ones, and the rest of the logic should work?

timriley:
This is now fixed by #266.

PR #266, opened 2023-11-23
group globs and mounts together

pat:
In the previous approach, globs would not be considered if a mounted app matched the route. On first glance this would be fair, but if that mounted app is at the root, and the glob is not, then even when the glob is listed first, it would be ignored (see #262).

But, a glob and a mount are not so different - they both have the potential to be far-reaching catch-all endpoints (though a glob is http_method-specific). So let’s handle them in the same collection, which helps keep their listed order in a routes file. This allows more-specific globs to be considered before a less-specific mount, provided they’re ordered that way in the routes.

This commit does not play so nicely with #263 - but the change there can be easily applied to MountedPath in this PR. Also, I realise there's potential for unit tests and documentation, but I would appreciate confirmation that this approach has merit first! :)

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -20,6 +20,8 @@ module Hanami
     require "hanami/router/block"
     require "hanami/router/route"
     require "hanami/router/url_helpers"
+    require "hanami/router/globbed_path"
+    require "hanami/router/mounted_path"
 
     # URL helpers for other Hanami integrations
     #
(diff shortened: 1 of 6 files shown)


timriley:
@pat I like the cut of your jib! Happy with this direction. Any chance you could fix the conflicts now that I've merged #263? And if you could throw in a few tests, then I reckon we can get this in :)

pat:
Thanks @timriley - I've just rebased and added some tests. RSpec and Rubocop are both happy for me locally, but I think CI's picked up a new Rubocop version and some new rules are failing things here.

pat:
Rebased to make Rubocop happy again :)

pat:
For some reason I thought there was already integration specs for this, but I see there are not. Give me a moment.

pat:
Thanks for the typo fix! 😁 - I'm feeling happy enough with all of this if you wish to merge it.

timriley:
This is a great improvement, thank you @pat!

PR #267, opened 2023-11-26
Pass through kwargs to middleware

pat:
Most middleware expect the old-school approach of hash options, but some are now using keyword arguments instead (such as Rack::Timeout). So let's presume that any keyword arguments that aren't known by Hanami should be passed through.

There is the edge-case of both the middleware and Hanami expecting arguments with the same name - this does not address that. 😓

This PR is paired with a forthcoming one for hanami/hanami, as that's where the various parameters are captured.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,3 +20,5 @@ Style/RescueStandardError:
 Style/CommentedKeyword:
   Exclude:
     - "spec/**/*.rb"
+Style/RedundantParentheses:
+  Enabled: false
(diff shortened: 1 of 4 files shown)


timriley:
Thanks for this, @pat! This change looks good to me (revisiting this code, it kind of perplexes me that this class is inside router, but only invoked from inside hanami, but that shall be a topic for another day).

The one thing I noticed when inspecting this is that the updated test would still pass even if I disabled the updated implementation, so I've pushed a new test that requires a keyword arg to be properly passed through to the middleware.

PR #273, opened 2024-09-04
Improve Trie implementation when different path variables used in same path location

### LINKED ISSUE #247: Failing edge cases for optional variables
I would like to use such pattern: `/(something/:optional/)and/:always` but router doesn't catch it. Failing test:

```ruby
    describe "optional variable with variable" do
      let(:router) do
        described_class.new do
          get "/(test1/:foo/)test2/:bar", as: :variable, to: RecognitionTestCase.endpoint("variable")
        end
      end

      it "recognizes route(s)" do
        runner.run!([
          [:variable, "/test1/foo/test2/bar", {foo: "foo", bar: "bar"}],
          [:variable, "/test2/bar", {foo: nil, bar: "bar"}]
        ])
      end
    end
```

>      Failure/Error: expect(name).to be_nil
>     
>       expected: nil
>            got: :variable

I'm not sure if it's bug or my bad usage?

### LINKED ISSUE #255: Routes with multiple variables not working in common nested route scenario
Hello Hanami team 👋 

First, thanks for the amazing project. It's a breeze to have projects like Hanami, dry-rb and ROM in the Ruby community 😄 

That being said, I've been using Hanami 2 lately I've came across the following issue with hanami-router:
```ruby
# Here I would expect the posts route to get matched

router = Hanami::Router.new do
  get "/users/:id", to: ->_ {[200, {}, ['show user']]}
  get "/users/:user_id/posts", to: ->_ {[200, {}, ['index posts']]}
end
router.call(Rack::MockRequest.env_for("/users/1/posts"))
# => [404, {"Content-Length"=>"9"}, ["Not Found"]]
```

Note, however, that this work fine:
```ruby
router = Hanami::Router.new do
  get "/users/:user_id", to: ->_ {[200, {}, ['show user']]}
  get "/users/:user_id/posts", to: ->_ {[200, {}, ['index posts']]}
end
router.call(Rack::MockRequest.env_for("/users/1/posts"))
# => [200, {}, ["index posts"]]
```

I would expect the response to be the same in both examples, regardless of the path variable name. Also worth noting, I think naming _the last_ "id" in the route generically as `id` (as opposed to always specifically calling it with the name of the resource, e.g. "user_id" or "post_id") is a common pattern that is also adopted in Rails by default.

Version used: 2.0.2

### LINKED ISSUE #256: Inconsistent trailing slash behaviour
Hello!

I've just found this another surprising behavior when writing a simple Hanami app, this time related to trailing slashes:

```ruby
router = Hanami::Router.new do
  get "/users", to: ->_ { [200, {}, ['index users']] }
  get "/users/:id", to: ->_ {[200, {}, ['show user']]}
end

# without trailing slash it works as expected
router.call(Rack::MockRequest.env_for("/users"))
# => [200, {}, ["index users"]]

# with trailing slash, I get a 404
router.call(Rack::MockRequest.env_for("/users/"))
# => [404, {"Content-Length"=>"9"}, ["Not Found"]]

# interestingly, the show route works just fine
router.call(Rack::MockRequest.env_for("/users/1/"))
=> [200, {}, ["show user"]]
```

Version used: 2.0.2

### LINKED ISSUE #272: Routes with suffixes conflict with each other
Given these two routes, the order in which the routes are declared in `config/routes.rb` impacts how the router behaves:

```ruby
get '/attendance/:season_slug/events', to: 'attendances.index', as: :select_event_for_attendance
get '/attendance/:event_slug', to: 'attendance.show', as: :attendances
```

Notice both routes have the same prefix. One route adds "/events" as a suffix.

When the routes are written as above, given the URL http://localhost:2300/attendance/qgox, I get a backtrace:

```
Hanami::Router::NotAllowedError at /attendance/qgox
Only DELETE, POST requests are allowed at /attendance/qgox

block in <module:ClassMethods>
hanami (2.2.0.beta1) lib/hanami/slice.rb, line 1072
Hanami::Slice::Router#not_allowed
hanami-router (2.2.0.beta1) lib/hanami/router.rb, line 640
Hanami::Slice::Router#call
hanami-router (2.2.0.beta1) lib/hanami/router.rb, line 106
Hanami::Middleware::Assets#call
rack (2.2.9) lib/rack/static.rb, line 161
Rack::Session::Cookie#context
rack (2.2.9) lib/rack/session/abstract/id.rb, line 266
Rack::Session::Cookie#call
rack (2.2.9) lib/rack/session/abstract/id.rb, line 260
Rack::MethodOverride#call
rack (2.2.9) lib/rack/method_override.rb, line 24
BetterErrors::Middleware#protected_app_call
better_errors (2.10.1) lib/better_errors/middleware.rb, line 87
BetterErrors::Middleware#better_errors_call
better_errors (2.10.1) lib/better_errors/middleware.rb, line 82
BetterErrors::Middleware#call
better_errors (2.10.1) lib/better_errors/middleware.rb, line 60
```

If I flip the order of the routes around, where `/events` is last, then routing occurs and I can render the page I was trying to visit. On the flip-side, trying to visit http://localhost:2300/attendance/abgf/events results in a different error:

```
Hanami::Router::NotFoundError at /attendance/abgf/events
No route found for GET /attendance/abgf/events

block in <module:ClassMethods>
hanami (2.2.0.beta1) lib/hanami/slice.rb, line 1077
Hanami::Slice::Router#not_found
hanami-router (2.2.0.beta1) lib/hanami/router.rb, line 646
Hanami::Slice::Router#call
hanami-router (2.2.0.beta1) lib/hanami/router.rb, line 106
Hanami::Middleware::Assets#call
rack (2.2.9) lib/rack/static.rb, line 161
Rack::Session::Cookie#context
rack (2.2.9) lib/rack/session/abstract/id.rb, line 266
Rack::Session::Cookie#call
rack (2.2.9) lib/rack/session/abstract/id.rb, line 260
Rack::MethodOverride#call
rack (2.2.9) lib/rack/method_override.rb, line 24
BetterErrors::Middleware#protected_app_call
better_errors (2.10.1) lib/better_errors/middleware.rb, line 87
BetterErrors::Middleware#better_errors_call
better_errors (2.10.1) lib/better_errors/middleware.rb, line 82
BetterErrors::Middleware#call
better_errors (2.10.1) lib/better_errors/middleware.rb, line 60
Hanami::Webconsole::Middleware#call
hanami-webconsole (2.2.0.beta1) lib/hanami/webconsole/middleware.rb, line 56
```

Full `config/routes.rb`:

```ruby
# frozen_string_literal: true

module Scoutatt
  class Routes < Hanami::Routes
    get '/seasons', to: 'seasons.index', as: :seasons
    get '/seasons/new', to: 'seasons.new', as: :new_season
    get '/seasons/:slug', to: 'seasons.show', as: :show_season
    post '/seasons', to: 'seasons.create', as: :create_season
    get '/seasons/:slug/edit', to: 'seasons.edit', as: :edit_season
    patch '/seasons/:slug', to: 'seasons.update', as: :update_season
    delete '/seasons/:slug', to: 'seasons.destroy', as: :destroy_season

    get '/seasons/:slug/events', to: 'events.index', as: :events
    post '/seasons/:slug/events', to: 'events.create', as: :create_event

    post '/registrations/:slug', to: 'registrations.create', as: :create_registrations

    post '/events', to: 'events.create'
    patch '/events/:slug', to: 'events.update', as: :update_event
    delete '/events/:slug', to: 'events.destroy', as: :destroy_event

    get '/attendance/:event_slug', to: 'attendance.show', as: :attendances
    get '/attendance/:season_slug/events', to: 'attendances.index', as: :select_event_for_attendance
    post '/attendance/registration/:registration_slug',
         to: 'attendance.create_from_registration',
         as: :create_attendance_from_registration
    post '/attendance/:event_slug',
         to: 'attendance.create', as: :create_attendance
    delete '/attendance/:registration_slug',
           to: 'attendance.destroy', as: :destroy_attendance
  end
end
```

### Environment

```
$ bundle show | ag hanami
  * hanami (2.2.0.beta1)
  * hanami-assets (2.2.0.beta1)
  * hanami-cli (2.2.0.beta1)
  * hanami-controller (2.2.0.beta1)
  * hanami-db (2.2.0.beta1)
  * hanami-reloader (2.2.0.beta1)
  * hanami-router (2.2.0.beta1)
  * hanami-rspec (2.2.0.beta1)
  * hanami-utils (2.2.0.beta1)
  * hanami-validations (2.2.0.beta1)
  * hanami-view (2.2.0.beta1)
  * hanami-webconsole (2.2.0.beta1)

$ ruby --version
ruby 3.2.4 (2024-04-23 revision af471c0e01) [arm64-darwin23]
```

dcr8898:
@kyleplump and I collaborated on this PR.

## Summary

Router cannot differentiate between multiple defined routes that contain different path variable names in the same location. For example:

```ruby
get "/:id/edit"
get "/:uuid"
```

This PR changes the router behavior to group all variable slugs together for matching purposes and to lazily create and evaluate Mustermann matchers. Discussion below.

The above symptom is illustrated in other open issues, such as #255 (Routes with multiple variables not working in common nested route scenario) and  #256 (Inconsistent trailing slash behaviour). Although trailing slashes are not a documented use case, they sometimes work (and sometimes don't) due to segment-by-segment matching.

The proposed implementation should close #255 , #256 , and #272 , by applying complete route matchers that provide consistent, predictable behavior, as documented in the current Hanami Guides.

We also suggest closing #247 (Failing edge cases for optional variables) as an undocumented use case that is not currently supported. Optional segments are not supported by either the current router or the proposed update.

Closes #247 
Closes #255
Closes #256 
Closes #272 

### Current Router Behavior

Hanami Router 2.0 uses a trie data structure to build a representation of all routes that contain a "dynamic segment," a.k.a., a path variable. It should be noted that all routes that contain only fixed segments are stored in a separate Hash where they can be looked up directly.

Notable features of the current router:

- Routes are stored in the trie as nodes based on each segment of the route (segments defined by the "/" delimiter in the route).
- For path variables/dynamic segments, a Mustermann matcher is created and stored in the trie node, with the matcher used as the key for the node.
- When matching an incoming path to a route, the trie evaluates each segment of the path and calculates and collects derived parameter values *from each dynamic segment as it goes* (this is important).

Because the Mustermann matcher for each dynamic segment is used as the key for that node, different matchers will result in different nodes being created, even if the dynamic segments in question are in parallel locations in different routes.

That is, in the route example above, different matchers would be created for `:id` and `:uuid`, since different path variable names are used.  However, the same node would be have been used for both routes if the same path variable name had been used (say, `:id`). It is worth noting also that different matchers (and therefore different nodes) would also be created if the same path variable names were used, but with different constraints.

The problem arises when matching paths and collecting parameters on the fly, as illustrated below.

The following routes:

```ruby
get "/:id/edit"
get "/:id"
```

Would produce a trie like this:

```mermaid
flowchart LR
  A((root))-->B(((":id")))
  B-->C((("edit")))
```

Where the double circles represent potential path endpoints, so both `:id` and `edit` are potential path endpoints.

In this case, the router has no trouble matching either `/1234` or `/1234/edit` to the correct routes.

However, if the following routes are used:

```ruby
get "/:id/edit"
get "/:uuid"
```

A trie like this will be created:

```mermaid
flowchart LR
  A((root))-->B((":id"))
  B-->C((("edit")))
  A-->D(((":uuid")))
```

Here, `:id` and `:uuid` are different nodes, and `:id` is no longer an endpoint. Now, when the router tries to match a path like "/9d5794e7-12ed-4f00-a6d1-a5be9cd4072e", it has no way to know that this is a UUID and incorrectly matches it to the `:id` node. Crucially, the UUID is collected and stored in the params Hash under the key `:id`. When the router then realizes that this path does not match any route at or below the `:id` node, it has no way to backtrack, "forget" the incorrect `:id` param, and search down a different node path. In fact, backtracking in this way is contrary to trie design and would essentially convert the trie to an ordinary tree data structure.

### Proposed Router Behavior

The proposed router behavior is to store a single variable node for all dynamic segments that are defined in parallel locations in an app's route table. Also, rather than attempting to parse and collect path variable params on the fly, the proposed router delays the creation and evaluation of Mustermann matchers until a potential full path/route match has been identified.

The following routes:

```ruby
get "/:id/edit"
get "/:uuid"
```

Would now create a trie like this:

```mermaid
flowchart LR
  A((root))-->B(((":*")))
  B-->C((("edit")))
```

And each endpoint would contain the information needed to create and evaluate a matcher for a proposed path that evaluates to that endpoint.

## Tests

The proposed implementation adds a new class, Leaf, and refactors the Trie and Node classes. New unit tests are included for each of these classes. New integration tests are also included (in `recognition_spec.rb`). All of these tests are passing.

The proposed implementation breaks two existing tests, [here](https://github.com/hanami/router/blob/e9c15474ec182a1094d933160c6635e7c0afe63c/spec/integration/hanami/router/recognition_spec.rb#L572-L584) and [here](https://github.com/hanami/router/blob/e9c15474ec182a1094d933160c6635e7c0afe63c/spec/integration/hanami/router/scope_spec.rb#L47-L61). These tests each break for reasons that merit further, separate discussion in another comment below, but we believe these tests exercise undocumented behavior and should be changed or deleted. In the alternative, if this behavior is desired, it should be documented and described and we would be happy to try to implement it accordingly.

Interestingly, there is also a commented-out test [here](https://github.com/hanami/router/blob/e9c15474ec182a1094d933160c6635e7c0afe63c/spec/integration/hanami/router/recognition_spec.rb#L552-L570) that documents the current issue, and now passes with the proposed changes. :nerd_face: 

## Conclusion

With guidance, we will delete or modify the breaking tests or, in the alternative, implement whatever documented behavior is desired for these use cases.

Finally, performance is important in this gem. We believe it is possible that the proposed implementation is faster than the current one, but we have not performance tested it. We have no reason to believe it will be slower. Memory usage is also not tested, but we believe fewer objects will be created with the proposed changes.

Thank you!

### DIFF
--- a/lib/hanami/router/leaf.rb
+++ b/lib/hanami/router/leaf.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require "mustermann/rails"
+
+module Hanami
+  class Router
+    class Leaf
+      # Trie Leaf
+      #
+      # @api private
+      # @since 2.1.1
+      attr_reader :to, :params
+
+      # @api private
+      # @since 2.1.1
+      def initialize(route, to, constraints)
+        @route = route
+        @to = to
+        @constraints = constraints
+        @params = nil
+      end
+
+      # @api private
+      # @since 2.1.1
+      def match(path)
+        match = matcher.match(path)
+
+        @params = match.named_captures if match
+
+        match
+      end
+
+      private
+
+      # @api private
+      # @since 2.1.1
+      def matcher
+        @matcher ||= Mustermann.new(@route, type: :rails, version: "5.0", capture: @constraints)
+      end
+    end
+  end
+end
--- a/lib/hanami/router/leaf.rb
+++ b/lib/hanami/router/leaf.rb
@@ -8,11 +8,11 @@ module Hanami
       # Trie Leaf
       #
       # @api private
-      # @since 2.1.1
+      # @since 2.2.0
       attr_reader :to, :params
 
       # @api private
-      # @since 2.1.1
+      # @since 2.2.0
       def initialize(route, to, constraints)
         @route = route
         @to = to
@@ -21,7 +21,7 @@ module Hanami
       end
 
       # @api private
-      # @since 2.1.1
+      # @since 2.2.0
       def match(path)
         match = matcher.match(path)
 
@@ -33,7 +33,7 @@ module Hanami
       private
 
       # @api private
-      # @since 2.1.1
+      # @since 2.2.0
       def matcher
         @matcher ||= Mustermann.new(@route, type: :rails, version: "5.0", capture: @constraints)
       end
--- a/lib/hanami/router/node.rb
+++ b/lib/hanami/router/node.rb
@@ -46,7 +46,7 @@ module Hanami
       end
 
       # @api private
-      # @since 2.1.1
+      # @since 2.2.0
       def match(path)
         @leaves&.find { |leaf| leaf.match(path) }
       end
--- a/lib/hanami/router/node.rb
+++ b/lib/hanami/router/node.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "hanami/router/segment"
+require "hanami/router/leaf"
 
 module Hanami
   class Router
@@ -18,15 +18,14 @@ module Hanami
       def initialize
         @variable = nil
         @fixed = nil
-        @to = nil
+        @leaves = nil
       end
 
       # @api private
       # @since 2.0.0
-      def put(segment, constraints)
+      def put(segment)
         if variable?(segment)
-          @variable ||= {}
-          @variable[segment_for(segment, constraints)] ||= self.class.new
+          @variable ||= self.class.new
         else
           @fixed ||= {}
           @fixed[segment] ||= self.class.new
@@ -35,36 +34,21 @@ module Hanami
 
       # @api private
       # @since 2.0.0
-      #
-      def get(segment) # rubocop:disable Metrics/PerceivedComplexity
-        return unless @variable || @fixed
-
-        found = nil
-        captured = nil
-
-        found = @fixed&.fetch(segment, nil)
-        return [found, nil] if found
-
-        @variable&.each do |matcher, node|
-          break if found
-
-          captured = matcher.match(segment)
-          found = node if captured
-        end
-
-        [found, captured&.named_captures]
+      def get(segment)
+        @fixed&.fetch(segment, nil) || @variable
       end
 
       # @api private
       # @since 2.0.0
-      def leaf?
-        @to
+      def leaf!(route, to, constraints)
+        @leaves ||= []
+        @leaves << Leaf.new(route, to, constraints)
       end
 
       # @api private
-      # @since 2.0.0
-      def leaf!(to)
-        @to = to
+      # @since 2.1.1
+      def match(path)
+        @leaves&.find { |leaf| leaf.match(path) }
       end
 
       private
@@ -73,18 +57,6 @@ module Hanami
       # @since 2.0.0
       def variable?(segment)
         Router::ROUTE_VARIABLE_MATCHER.match?(segment)
-      end
-
-      # @api private
-      # @since 2.0.0
-      def segment_for(segment, constraints)
-        Segment.fabricate(segment, **constraints)
-      end
-
-      # @api private
-      # @since 2.0.0
-      def fixed?(matcher)
-        matcher.names.empty?
       end
     end
   end
--- a/lib/hanami/router/trie.rb
+++ b/lib/hanami/router/trie.rb
@@ -21,33 +21,26 @@ module Hanami
 
       # @api private
       # @since 2.0.0
-      def add(path, to, constraints)
+      def add(route, to, constraints)
+        segments = split(route)
         node = @root
-        for_each_segment(path) do |segment|
-          node = node.put(segment, constraints)
+
+        segments.each do |segment|
+          node = node.put(segment)
         end
 
-        node.leaf!(to)
+        node.leaf!(route, to, constraints)
       end
 
       # @api private
       # @since 2.0.0
       def find(path)
+        segments = split(path)
         node = @root
-        params = {}
 
-        for_each_segment(path) do |segment|
-          break unless node
+        return unless segments.all? { |segment| node = node.get(segment) }
 
-          child, captures = node.get(segment)
-          params.merge!(captures) if captures
-
-          node = child
-        end
-
-        return [node.to, params] if node&.leaf?
-
-        nil
+        node.match(path)&.then { |found| [found.to, found.params] }
       end
 
       private
@@ -58,10 +51,11 @@ module Hanami
       private_constant :SEGMENT_SEPARATOR
 
       # @api private
-      # @since 2.0.0
-      def for_each_segment(path, &blk)
+      # @since 2.1.1
+      def split(path)
         _, *segments = path.split(SEGMENT_SEPARATOR)
-        segments.each(&blk)
+
+        segments
       end
     end
   end
--- a/lib/hanami/router/trie.rb
+++ b/lib/hanami/router/trie.rb
@@ -22,7 +22,7 @@ module Hanami
       # @api private
       # @since 2.0.0
       def add(route, to, constraints)
-        segments = split(route)
+        segments = segments_from(route)
         node = @root
 
         segments.each do |segment|
@@ -35,7 +35,7 @@ module Hanami
       # @api private
       # @since 2.0.0
       def find(path)
-        segments = split(path)
+        segments = segments_from(path)
         node = @root
 
         return unless segments.all? { |segment| node = node.get(segment) }
@@ -52,7 +52,7 @@ module Hanami
 
       # @api private
       # @since 2.1.1
-      def split(path)
+      def segments_from(path)
         _, *segments = path.split(SEGMENT_SEPARATOR)
 
         segments
--- a/lib/hanami/router/trie.rb
+++ b/lib/hanami/router/trie.rb
@@ -51,7 +51,7 @@ module Hanami
       private_constant :SEGMENT_SEPARATOR
 
       # @api private
-      # @since 2.1.1
+      # @since 2.2.0
       def segments_from(path)
         _, *segments = path.split(SEGMENT_SEPARATOR)
 
--- a/spec/integration/hanami/router/recognition_spec.rb
+++ b/spec/integration/hanami/router/recognition_spec.rb
@@ -601,20 +601,6 @@ RSpec.describe Hanami::Router do
       end
     end
 
-    describe "relative variable with permissive constraint" do
-      let(:router) do
-        described_class.new do
-          get ":test", as: :regex, test: /.*/, to: RecognitionTestCase.endpoint("regex")
-        end
-      end
-
-      it "recognizes route(s)" do
-        runner.run!([
-          [:regex, "/test/", {test: "test"}]
-        ])
-      end
-    end
-
     describe "variable with permissive constraint" do
       let(:router) do
         described_class.new do
--- a/spec/integration/hanami/router/recognition_spec.rb
+++ b/spec/integration/hanami/router/recognition_spec.rb
@@ -256,11 +256,27 @@ RSpec.describe Hanami::Router do
       end
     end
 
-    describe "variable and variable with fixed" do
+    describe "variable followed by variable with fixed with different slug names" do
       let(:router) do
         described_class.new do
           get "/:foo",     as: :variable_one, to: RecognitionTestCase.endpoint("variable_one")
           get "/:bar/baz", as: :variable_two, to: RecognitionTestCase.endpoint("variable_two")
+        end
+      end
+
+      it "recognizes route(s)" do
+        runner.run!([
+          [:variable_one, "/one", {foo: "one"}],
+          [:variable_two, "/two/baz", {bar: "two"}],
+        ])
+      end
+    end
+
+    describe "variable with fixed followed by variable with different slug names" do
+      let(:router) do
+        described_class.new do
+          get "/:bar/baz", as: :variable_two, to: RecognitionTestCase.endpoint("variable_two")
+          get "/:foo",     as: :variable_one, to: RecognitionTestCase.endpoint("variable_one")
         end
       end
 
--- a/spec/integration/hanami/router/recognition_spec.rb
+++ b/spec/integration/hanami/router/recognition_spec.rb
@@ -256,6 +256,22 @@ RSpec.describe Hanami::Router do
       end
     end
 
+    describe "variable and variable with fixed" do
+      let(:router) do
+        described_class.new do
+          get "/:foo",     as: :variable_one, to: RecognitionTestCase.endpoint("variable_one")
+          get "/:bar/baz", as: :variable_two, to: RecognitionTestCase.endpoint("variable_two")
+        end
+      end
+
+      it "recognizes route(s)" do
+        runner.run!([
+          [:variable_one, "/one", {foo: "one"}],
+          [:variable_two, "/two/baz", {bar: "two"}],
+        ])
+      end
+    end
+
     describe "relative variable with constraints" do
       let(:router) do
         described_class.new do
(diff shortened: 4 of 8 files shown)


adam12:
This is a fantastic PR :clap:

I looked at this issue a couple years ago and wasn't able to solve it due to time constraints. I originally tried to keep a stack of each opportunity to branch and then just keep trying each branch from the stack when there was a routing miss. I like your solution a lot better.

Well done! 🥂

dcr8898:
@adam12 Thanks!

We considered using a search algorithm like that, but in addition to the speed issue, you would have to maintain a parallel stack with the params collected on-the-fly (and prune it as needed). That was starting to feel complicated, so we looked for different solutions that would let us keep the speed a trie offers.

Cheers!

dcr8898:
I was looking over the first test discussed above that is broken by the current proposal:

```ruby
    describe "relative variable with permissive constraint" do
      let(:router) do
        described_class.new do
          get ":test", as: :regex, test: /.*/, to: RecognitionTestCase.endpoint("regex")
        end
      end


      it "recognizes route(s)" do
        runner.run!([
          [:regex, "/test/", {test: "test"}]
        ])
      end
    end
```

I noticed that the initial route in this test is also set up in a non-standard (that is, undocumented) way: there is no leading `/`. I think this is a symptom of the practice of using Mustermann matchers in a piecemeal manner on individual path segments: it introduces edge cases and inconsistent behavior by never applying the complete route specification from `routes.rb`.

This problem is illustrated in other open issues, such as #256 (Inconsistent trailing slash behaviour). Although trailing slashes are not a documented use case, they sometimes work (and sometimes don't) due to segment-by-segment matching.

The proposed implementation should close #255, close #273, and close #256, by applying complete route matchers that impose consistent, predictable behavior, as documented in the current Hanami Guides.

I would also suggest closing #247 (Failing edge cases for optional variables) as an undocumented use case that is not currently supported. Optional arguments are not supported by either the current router or the proposed update.

dcr8898:
I see that the CI build is failing at the RSpec step. I don't think this is caused by anything in this pull request.

timriley:
Thanks for this amazing PR, folks! And I think it's so cool that the two of you worked together on this 😍 

I want to give this proper attention, so it'll need some brain space. Right now I'm finishing off multiple gateway support for Hanami's DB layer. As soon as I'm done with that, attending to this PR will be my #1 priority. Speak again soon :)

timriley:
I haven't forgotten this, folks! I'm releasing rc1 tomorrow and I'll make sure this is reviewed/merged before 2.2.0.

dcr8898:
Let us know if you would like to discuss at all. Questions and explanations might be faster that way. Thanks!

timriley:
Update: I've started looking at this!

timriley (commented):
OK! I'm here! Let's do this.

Firstly, thank you again for your work here, @dcr8898 and @kyleplump! This is a stunningly well put together PR! 🤩

Your explanation and diagrams were extremely helpful for me to understand the issue and the changes you made, and I think your code changes make for a much better overall implementation.

I'd love to see this merged! 💪🏼 

(In fact, I'd also love it if either one of you felt like writing a technical overview document to add to this repo, or peppering these classes with class- and method-level documentation comments explaining how they work. As you no doubt have discovered, it can take a bit of time to understand how everything hangs together, and the more we can do to make that easier for others and our future selves, the better! At this point you two are probably among the foremost experts in this codebase. Let me know what you think!)

Now let me address the matter of the two failing specs:

## First Broken Test

> The first broken test is [here](https://github.com/hanami/router/blob/e9c15474ec182a1094d933160c6635e7c0afe63c/spec/integration/hanami/router/recognition_spec.rb#L572-L584), in `integration/hanami/router/recognition_spec.rb`.

I agree with your recommendations for this. All our public documentation for routes (https://guides.hanamirb.org/v2.1/routing/overview/) shows leading slashes in all cases. That's the behaviour we care about. Let's delete the test and document the change in behaviour.

## Second Broken Test

> The second breaking test is [here](https://github.com/hanami/router/blob/e9c15474ec182a1094d933160c6635e7c0afe63c/spec/integration/hanami/router/scope_spec.rb#L47-L61), in `integration/hanami/router/scope_spec.rb`.

You know what, I think the current passes/failures within this test are what I want! This behaviour feels right to me, and I would be inclined to remove the one failing test case for `GET /it/` (line 59) to allow this test to pass again.

Let me explain my thinking, keeping things mostly at a user perspective.

My first thought relates to the other broken test, above: we do not provide any automatic handling of trailing slashes in Hanami routes. If you want a trailing slash in a route, you have to provide one. Given this, why should we match on the request for `/it/` here? That request has a trailing slash, and we haven't specified one in the routes.

Next, if we look at scopes in general, [our routing guide](https://guides.hanamirb.org/v2.1/routing/overview/) shows them working without the user having to supply any slashes:

```
    scope "about" do
      get "/contact-us", to: "content.contact_us" # => /about/contact-us
      get "/faq", to: "content.faq" # => /about/faq
    end
```

Our tests show a similar story. `spec/integration/hanami/router/scope_spec.rb` has these routes:

```
    scope "trees" do
      root               to: ->(*) { [200, {}, ["Trees (GET)!"]] }
      get     "/cherry", to: ->(*) { [200, {}, ["Cherry (GET)!"]] }
      # ...
    end
```

Leading to these successful requests:

```
  expect(app.request("GET", "/trees", lint: true).body).to eq("Trees (GET)!")
  expect(app.request("GET", "/trees/cherry", lint: true).body).to eq("Cherry (GET)!")
```

From these examples, the way I see it is that using `scope` buys you a URL segment with an implicit leading slash.

From there, our existing documented behaviour is that the root of a scope should match that exact scope with no trailing slash. This is how it works for static scopes now (`scope "trees" do`), and that's what our tests are also showing for a scope with a dynamic segment (`scope ":locale" do`), with the exception of the now-failing `GET /it/` somehow having worked under the prior implementation.

So like the first failing test above, I would chalk this up to undocumented behaviour simply happening to work by coincidence. I think we should drop the one offending line from this test and document the change.

---

The upshot of the decisions here would be that we can very easily move forward with this PR. Delete the first test and delete a single line from this second test and we are good to go!

What do you think?

(I have broader thoughts about how we could make our routes nicer, i.e. doing away with requiring leading slashes altogether, thereby unifying normal routes and scopes, but those are matters for another day!)

If you're happy with the decisions here, I'm happy to make the changes required to this PR and get it merged (though if you want to beat me to it, that's fine too).

I'll also look at the rexml thing separately and sort that out.

Thanks again for your marvellous work!

timriley:
@dcr8898 @kyleplump rexml issue is fixed and main branch builds are green and running again.

Part of this fix was https://github.com/hanami/router/commit/d2143c2481027dc1c9f19e59662d6f4bb08dd492, which introduced a merge conflict with this PR. Hopefully a simple one to address though: just dealing with a single "useless assignment" line that I deleted in node.rb.

dcr8898:
Thank you, Tim! I'd like to confer with @kyleplump before replying. We should get back to you shortly!

timriley:
Thanks for the update, @dcr8898. One way or another, I _do_ want to get this PR included in 2.2, since it's an important bug fix.

The release is planned for 36 hours from now. I'll hold this PR until the very last minute, if need be, but of course I'd appreciate any feedback from you in the meantime :)

Once again, I'm sorry for the tight timeline here.

dcr8898:
I spoke to @kyleplump before implement the following.

## Changes to tests

- We deleted the failing test from recognition_spec for the reasons discussed.
- I thought I disagreed with you regarding the second failing test in scope_spec, but  I realized I was wrong. :nerd_face:  Therefore, we deleted the single failing expectation in this test.
- Consistent with my separate response above, we removed unneeded tests of duplicate behavior, as you pointed out and recommended.

## Other changes
- Bumped yardoc version references to "2.2.0" for new classes/methods, consistent with your plans for release.
- The conflict resolution is trivial. I will do that in Github shortly.

## Other issues of note

The middleware module in this gem contains a parallel implementation of the routing trie. I did not investigate or make any changes to this module. It probably needs to be refactored in parallel with the changes to the router module.

You mentioned a technical overview doc or comprehensive yardocs. Do you have a preference? If a technical doc, what would that look like? A more comprehensive readme or something else?

Thank you! This was an intimidating project (for me). Kyle's willingness to dive right in helped me tremendously. Once we dug in, it was fun to figure out!

Thank you all! :purple_heart:

timriley:
Love it, thank you so much @dcr8898 and @kyleplump! Merging this and getting on with release things tonight. Will follow up about technical docs, etc. at a future time :)

PR #277, opened 2024-12-31
Rack 3 support

kyleplump:
multiple gems are impacted and updated by this change. this gem is the most significantly changed. i'll link to the open PR's of the affected gems below.  please let me know if you'd like to see any changes, thanks!

## highlights
changelog for Rack 3: https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md

- response headers are now [required to be lower case](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#response-headers-must-be-lower-case)
- `rack.input` is now [no longer rewindable](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#rackinput-is-no-longer-required-to-be-rewindable)
    - this impacts [processing multiform data](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#rackinput-is-no-longer-rewound-after-consuming-form-and-multipart-data) as well

- some specs needed updating, on account that the `content-length` response header no longer a part of the implementation of `Rack::Response`. this change [happened in v2.1](https://github.com/rack/rack/issues/1472#issuecomment-574362342), and while not directly related to Rack 3, was included here to get all green

## impacted gems
- hanami: https://github.com/hanami/hanami/pull/1493
- cli: https://github.com/hanami/cli/pull/289
- controller: https://github.com/hanami/controller/pull/460
- devtools: https://github.com/hanami/devtools/pull/90


cllns:
Excellent! Thanks so much for doing this. It'll be a huge win when we can make a release that supports Rack 3.

Looks like there are some Rubocop errors, can you fix those?

I have a few questions (for you and others):
* Do we want to _require_ rack 3 or simply support it, if people want to upgrade themselves?
  * If we require rack 3, then we should add support for people who have mixed-cased headers, so it's not a breaking change. It seems like [Rack::Headers does this](https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#response-headers-must-be-lower-case).
  * If we don't require rack 3, then we might have to add conditionals based on rack version. Hopefully we can avoid this or at least minimize it.
* Do we _need_ our input to be rewindable with `Rack::RewindableInput::Middleware` at all? What if we just consume the headers / form data as part of our parsing, add it to the env/params, then never need to rewind? Is that an option? I guess it's more performant.
* Similarly, for Content-Length, do we want add `Rack::ContentLength` middleware to all hanami-router apps, by default?
* Along those lines, if we need rewindable input for normal non-streaming HTTP data, then can we let people opt out so they can use Rack 3's streaming support? This could definitely be a separate PR that comes later.

kyleplump:
thanks for checking this out @cllns ! 

 fixed the rubocop offenses

- this is an excellent point. i feel as though the better DX would be to _support_ and not _require_ rack 3.  the number of code changes isn't so significant that i think we can get away with only a few conditionals, though I'll defer to your judgement here
- i'm not sure we _need_ rewindable inputs. i think your suggestion of putting it into the env/params would certainly work (and already has some precedent, re: `Router#_params`). I kept the rewindable input just to match what was already there, but would be happy to do away with it if we want to avoid bringing in another middleware
- i think having the `Rack::ContentLength` middleware is a great idea, and can add this in
- another excellent point. I'd love to provide more 'rack 3+ specific' features, and will definitely start thinking about how to include that into the existing gem.  including them here, or in a separate PR, i'll once again defer to your excellent judgement 😄

cllns:
> * this is an excellent point. i feel as though the better DX would be to _support_ and not _require_ rack 3.  the number of code changes isn't so significant that i think we can get away with only a few conditionals, though I'll defer to your judgement here

Sweet! If it's possible, I think that would be best, so we can keep this a minor version release. 

We'll have to add rack-version to our CI matrix, but that's not a problem. It makes sense for a router.

Now that I think about it, even if Rack 3 is optional, we don't want users who are setting mixed-case headers to break their apps, so we should use Rack::Headers regardless.

> * i'm not sure we _need_ rewindable inputs. i think your suggestion of putting it into the env/params would certainly work (and already has some precedent, re: `Router#_params`). I kept the rewindable input just to match what was already there, but would be happy to do away with it if we want to avoid bringing in another middleware

Outside of whether we _need_ it, I feel like there might be reasons we _want_ rewindable input... I guess it's more performant to keep the same input & rewind it instead of copying everything we need over. Plus users may be rewinding already (e.g. in middleware) and we don't want that to break.

Curious how others feel about this. I think it might make sense to use rewindable input by default for Rack 3 (maybe based on content type?) but hopefully there's a way to automatically allow non-rewindable inputs as well, for streams. If it can't be automatic then we should try to make it simple to opt-out of rewindable inputs.

I wish I knew more about Rack. I've read the UPGRADE_GUIDE, and started 'watching' the repo a couple weeks ago, so hopefully I can get a better grasp on it.

> * i think having the `Rack::ContentLength` middleware is a great idea, and can add this in

Looking into it, the [Rack::ContentLength middleware](https://github.com/rack/rack/blob/main/lib/rack/content_length.rb) only applies to responses that don't have the header already set. What if we just set it in hanami/hanami and leave it out of the router? Then we can document how users can add it. This keeps the router as small and flexible as possible. I think it's probably better to set it ourselves in the framework than rely on the middleware (which seems like more of a fallback).


> * another excellent point. I'd love to provide more 'rack 3+ specific' features, and will definitely start thinking about how to include that into the existing gem.  including them here, or in a separate PR, i'll once again defer to your excellent judgement 😄

This is already a big change (across several repos, thank you!), so I think it'd be wise to keep the scope of this as small as we can. Then we can work on Rack 3+ features, and ship those independently.

So I think at this point:
* Relax requirement to Rack > 2
* Add testing Rack 2 and Rack 3 to CI matrix for this gem (and hanami/hanami?)
* Make sure we support both, using RewindableInput
* Un-do the changes for updating the specs to lowercase headers, to ensure we don't break people using mixed case headers.

Then we can get these merged, and work on non-rewindable inputs in separate PR's. We could ship rewindable-only Rack 3 support or wait until we actually support non-rewindable input to ship a new versions of the gems. 

Curious to get @timriley's feedback :)

kyleplump:
hi @cllns 👋 (sorry its been 10 years)

i just committed some changes. some thoughts:

- relaxed the rack requirement
- allowed handling of mixed case headers
- injected content-length when necessary, not relying on the middleware

---
### headers
this was the trickiest part for me, and would be very curious to hear your thoughts on it.  i ran into a few different roadblocks, but think i have a solution:

- all of the specs had hard coded, mixed case headers to assert against.  in the previous iteration of this PR, i just lower cased them to match Rack3 spec.  in this new iteration, I left the headers untouched _but_ wrapped them in a new spec helper which either passes them through as is (for < rack 3), or relies on `Rack::Headers` for > rack 3.
- http headers that the router automatically injects (e.g., sometimes `location`, sometimes `allow`), are both 'checking' rack version (via a check for `Rack::Headers`), and responding appropriately.  Casing is incompatible between versions: `Allow` breaks Rack 3, but is needed for < Rack 3, etc.

these changes have achieved our own passing tests, as well as allowing the router to be able to respond correctly to multiple, technically incompatible, versions of Rack with just a few conditionals

### content-length
I'm honestly slightly confused by what you mean here.  i feel like automatically calculating and injecting the `content-length` header is the perfect work for the router? I added that behavior here, but would be happy to pull it out if you feel its better placed elsewhere.  

---
i'd love to keep cracking at this and want to make sure its right, so let me know any changes you'd like to see. and thank you again!

lib/hanami/router.rb
```diff
@@ -916,15 +922,18 @@ def _redirect(to, code)
     def _params(env, params)
       params ||= {}
       env[PARAMS] ||= {}
+      input = Rack::RewindableInput.new(env[::Rack::RACK_INPUT]) if env[::Rack::RACK_INPUT]
```
timriley, line 925: I wonder if we're "burying the lede" a little bit by doing this conversion of the input to a rewindable input down here in this `_params`.

The job of the `_params` method is to _extract params_, and by adding this line, I feel like we're asking it to do more than that, which isn't a good design change.

If we're deciding that we want to force a rack 3 input to be rewindable as the first part of our rack 3 transition, couldn't we do this conversion right up in `Router#call`?

`Router#call` is is the main entry point for the router, and as such, it feels like a more appropriate place to do something like this. It would make the rewindable conversion more prominent, and give us a clearer place to comment why we're doing it (this is important, IMO).

Can you please have a go at moving it there?

And as part of moving it there, I wonder if we actually want to wrap it in the rack_3? conditional, e.g.

```ruby
# Explanatory comment goes here
if rack_3? && (input = env[::Rack::RACK_INPUT])
  env[::Rack::RACK_INPUT] = Rack::RewindableInput.new(input)
end
```

Having this lead with the `rack_3?` conditional would also make it clearer that it's part of our Rack 3 transition strategy.

spec/support/http.rb
```diff
@@ -18,6 +18,20 @@ def self.testable?(mounted, requested)
         (mounted == requested) ||
           (mounted == "get" && requested == "head")
       end
+
+      def self.headers(expected)
```
timriley, line 22:
```suggestion
      def self.headers(headers_hash)
```

Having "expected" as an arg name here feels a bit confusing because that's a name we typically use when making test assertions/expectations. Here we're just conditionally converting the a headers hash into a different kind of hash-like object. So I think it's best to avoid the misleading argument name.

timriley: It would also be good if we added a line of docs above this method, explaining why we have it and what it does.

spec/unit/hanami/router/redirect_spec.rb
```diff
@@ -12,8 +12,14 @@
       env = Rack::MockRequest.env_for("/redirect")
       status, headers, = router.call(env)
 
+      location_header = if Hanami::Router.modern_rack?
+                          headers.fetch("location")
```
timriley, line 16:
```suggestion
                          headers["location"]
```

Having the `.fetch` on one side of the conditional and not the other imparts special emphasis on the fetch, which I don't think is warranted here. Even if `headers` don't have a `"location"` key (which would fail the fetch), if we used `headers["location"]` instead, we'd get a `nil` and still end up with a failing expectation a few lines below, which is a more useful outcome from the test than a `KeyError`.

(This suggestion applies to here as well as the few other places in our specs where we're doing a similar thing)

timriley: While we're here, let's use sensible indentation 😅 

```ruby
headers =
  if blah?
    "blorp"
  else
    "fneep"
  end
```

I'm happy with basically anything except the indentation style you've used here 😆

spec/support/http.rb
timriley: Another thought: I'm not in love with the quite verbose full `RSpec::Support::HTTP.headers` method name we have to pepper all over our spec files.

What if we:

1. moved this method onto another module, say `RSpec::Support::Rack`
2. called the method `rack_headers`
3. and then did an RSpec `config.include RSpec::Support::Rack`
4. allowing us to do a plain `rack_headers(...)` method call throughout our specs?

That feels like a lot less noise in the tests, helping us keep the focus on the actual header values that we're passing to the method.

Would you be up for giving that a go?

timriley (commented):
Thank you so much for doing this, @kyleplump, and thank you @cllns for the wonderful feedback on this PR so far! 😍 

I agree that we should be supporting both Rack 2 and 3 for the time being. Hanami 3 (whenever that comes) would be the right time to drop Rack 2 support.

If I might attempt summarise the changes here, adding Rack 3 compatibility boils down to:

1. Loosening the gem dependency (of course)
2. Deciding that we'll keep rewindable input objects as a first step of this transition, since that's what we've been expecting with Rack 2 to date
3. Updating our tests to expect the lower-cased headers that Rack 3 gives us

That's basically it, right?

If so, that's a really nice and simple adjustment. Easy to communicate, easy (hopefully) for others to understand. And it means the only thing we have to focus on for a full Rack 3 switch is a decision about rewindable input.

Love your work!

As for this PR, I've made a range of small suggestions via inline comments.

Biggest one is about moving the work of converting the input to rewindable up into `Router#call`, since I think that is an important behaviour, and keeping it buried in `Router#_params` makes it harder for us and our future contributors to find and understand.

Have a read of it all and please let know if there's anything else you'd like to discuss!

lib/hanami/router/block.rb
```diff
@@ -66,6 +66,10 @@ def params
         # @since 2.0.0
         def call
           body = instance_exec(&@blk)
+
+          # explicitly set 'content-length' header
+          headers[::Rack::CONTENT_LENGTH] = body.size.to_s
```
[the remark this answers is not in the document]
timriley: Sorry, I missed the detail about Content-Length from https://github.com/rack/rack/issues/1472#issuecomment-574362342. Now I'm now caught up.

I think we should leave `Content-Length` out. People on the latest version of Rack 2 are already missing this header anyway.

If we wanted to restore it, I think we could consider adding the `Rack::ContentLength` middleware as part of the default middleware stack in Hanami apps.

People using the router directly can choose to do the same, too. I think it's better to keep the router as lean as practical.

lib/hanami/middleware/body_parser.rb
```diff
@@ -39,10 +39,10 @@ def initialize(app, parsers)
       end
 
       def call(env)
-        body = env[RACK_INPUT].read
-        return @app.call(env) if body.empty?
+        body = env[RACK_INPUT]&.read
```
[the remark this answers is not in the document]
kyleplump: its actually to fix a failing [spec](https://github.com/hanami/router/blob/fe925ecadd0ad3633e8b674c7f87164fd1670acf/spec/integration/hanami/router/body_parsing_middleware_spec.rb#L26)

i figure if its failing in our existing test suite, that its chance of occurring for others is non-negligible and would be an extra safe guard.  happy to remove and rework the spec if you'd rather take that path!

lib/hanami/router/block.rb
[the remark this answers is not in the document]
kyleplump: My understanding is its actually the opposite: Rack 2 were the only people _getting_ the header on the rack responses.  The inclusion here was for green specs, but am super willing to either remove the explicit header checks from the specs, or go the middleware route

spec/support/http.rb
kyleplump: great suggestion, happy to make the change

spec/unit/hanami/router/redirect_spec.rb
kyleplump: _lol_ point taken! i'll give change it to my second favorite indentation style 😉

kyleplump (commented):
thanks for the review @timriley! i at least responded to all your comments / committed the suggestions. I have a few slightly larger updates on my end and will ping you back!

kyleplump:
@timriley pinging you back. made a bunch of requested updates, and left a few comments 🙂

please let me know what you think. thanks!

lib/hanami/router.rb
```diff
@@ -916,15 +922,18 @@ def _redirect(to, code)
...
 
-      if !env.key?(ROUTER_PARSED_BODY) && (input = env[::Rack::RACK_INPUT]) and input.rewind
+      if !env.key?(ROUTER_PARSED_BODY) && input
```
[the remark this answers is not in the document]
timriley: Cool :) Thanks for bringing it back. I just pushed up d211b91d82c9559100cd6df7f137b8b2f932ffa9, which restores this exact line of code as before.

lib/hanami/middleware/body_parser.rb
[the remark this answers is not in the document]
timriley: Cool, let's leave this in, then :)

lib/hanami/middleware/body_parser/form_parser.rb
```diff
@@ -30,6 +30,7 @@ def self.mime_types
         # @since 2.0.1
         # @api private
         def parse(*, env)
+          env[::Rack::RACK_INPUT].rewind if env[::Rack::RACK_INPUT].respond_to?(:rewind)
```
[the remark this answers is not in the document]
timriley: I figured out the reason for this.

Rack 2's `Rack::Multipart.parse_multipart` would do a rewind internally.

In Rack 3, this no longer happens. So for this reason we need to bring the rewinding into the hanami-router layer, at least until such time that we abandon the backwards-compatibility position of forcing the input to be rewindable.

I pushed f7116c8cbaefeac610dfd66b75ecfad9fc5654fc with a comment explaining why.

timriley (commented):
Everything is looking good here, @kyleplump! All that's left for us to figure out is the ContentLength issue. Which I will come back to ASAP, but I'm submitting the rest of this review in the meantime :)

cramt:
🙌🙌🙌🙌🙌

timriley:
@Andsbf Yes, we'll release this as soon as we can. Rack 3 support will be the main feature of an upcoming Hanami 2.3 release. It's not quite ready yet, there are still some incompatibilities I need to iron out on the main branch of hanami-controller.

PR #281, opened 2025-02-14
Convert dash (and more) to underscore in URL helper names

### LINKED ISSUE #280: Convert dashes in scope/slice URL fragments to underscores in URL helpers
By convention (better SEO etc), dashes are preferred over underscores in URLs, but this results in ugly helpers. Here's an example:

```ruby
# config/routes.rb

module Dashy
  class Routes < Hanami::Routes

    get "foo", to: "action.foo", as: :foo

    scope "myscope" do
      get "foo", to: "action.foo", as: :foo
    end

    scope "my-scope" do
      get "foo", to: "action.foo", as: :foo
    end

    slice :myslice, at: "/myslice" do
      get "foo", to: "action.foo", as: :foo
    end

    slice :my_slice, at: "/my-slice" do
      get "foo", to: "action.foo", as: :foo
    end
  end
end
```

The following routes and URL helpers are created:

```
GET  /foo            action.foo   as :foo
GET  /myscope/foo    action.foo   as :myscope_foo
GET  /my-scope/foo   action.foo   as :"my-scope_foo"  <--
GET  /myslice/foo    action.foo   as :myslice_foo
GET  /my-slice/foo   action.foo   as :"my-slice_foo"  <--
```

For the URL helpers, the dashes should be converted to underscores like so:

```
GET  /foo            action.foo   as :foo
GET  /myscope/foo    action.foo   as :myscope_foo
GET  /my-scope/foo   action.foo   as :my_scope_foo   <--
GET  /myslice/foo    action.foo   as :myslice_foo
GET  /my-slice/foo   action.foo   as :my_slice_foo   <--
```

Discussion: https://discourse.hanamirb.org/t/route-scopes-with-dashes/1120

svoop:
Given a scope (or by extension: a slice) which contains dash `-`, plus `+`, tilde `~` or dot `.` in their URL fragment:

```ruby
# config/routes.rb

module Dashy
  class Routes < Hanami::Routes
    scope "my-scope" do
      get "foo", to: "action.foo", as: :foo
    end
  end
end
```

The dashes etc. are converted to underscores for the corresponding URL helper name:

```
% hanami routes

GET   /my-slice/foo   action.foo   as :my_slice_foo
        ^^^                             ^^^
```

Closes #280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Rack compatible HTTP router for Ruby
 
+## [Unreleased]
+
+### Changed
+
+- [Sven Schwyn] Convert -, +, ~, and . to underscore in URL helper names (#280)
+
 ## v2.2.0 - 2024-11-05
 
 ### Changed
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -693,6 +693,10 @@ module Hanami
     # @api private
     PREFIXED_NAME_SEPARATOR = "_"
 
+    # @since x.x.x
+    # @api private
+    UNDERSCORED_NAME_REGEXP = /[\-+~.]/
+
     # @since 2.0.0
     # @api private
     ROOT_PATH = "/"
@@ -800,7 +804,7 @@ module Hanami
       end
 
       if as
-        as = prefixed_name(as)
+        as = prefixed_underscored_name(as)
         add_named_route(path, as, constraints)
       end
 
@@ -872,10 +876,13 @@ module Hanami
       @path_prefix.join(path).to_s
     end
 
-    # @since 2.0.0
+    # @since x.x.x
     # @api private
-    def prefixed_name(name)
-      @name_prefix.relative_join(name, PREFIXED_NAME_SEPARATOR).to_sym
+    def prefixed_underscored_name(name)
+      @name_prefix
+        .relative_join(name, PREFIXED_NAME_SEPARATOR)
+        .gsub(UNDERSCORED_NAME_REGEXP, "_")
+        .to_sym
     end
 
     # Returns a new instance of Hanami::Router with the modified options.
--- a/spec/unit/hanami/router/url_spec.rb
+++ b/spec/unit/hanami/router/url_spec.rb
@@ -9,6 +9,9 @@ RSpec.describe Hanami::Router do
       get "/books/:id", id: /\d+/, to: e, as: :constraints
       get "/articles(.:format)",   to: e, as: :optional
       get "/files/*glob",          to: e, as: :glob
+      scope "/a-b+c~d.e" do
+        get "/hanami", to: e, as: :scoped
+      end
     end
   end
 
@@ -45,6 +48,10 @@ RSpec.describe Hanami::Router do
 
     it "escapes additional params in query string" do
       expect(router.url(:fixed, return_to: "/dashboard")).to eq(URI("#{base_url}/hanami?return_to=%2Fdashboard"))
+    end
+
+    it "prefixes the scope using underscores" do
+      expect(router.url(:a_b_c_d_e_scoped)).to eq(URI("#{base_url}/a-b+c~d.e/hanami"))
     end
 
     # FIXME: should preserve this behavior?


lib/hanami/router.rb
```diff
@@ -693,6 +693,10 @@ def env_for(env, params = {}, options = {})
     # @api private
     PREFIXED_NAME_SEPARATOR = "_"
 
+    # @since 2.3.0
+    # @api private
+    NAME_UNDERSCORE_REGEXP = /[\-+~.]/.freeze
```
cllns, line 698: I don't think the leading `\` does anything, since it just acts as an escape, and `-` doesn't need to be escaped at the beginning of a regexp. It only acts as a regexp separator when it's between two characters. I'm also fine with keeping it though, just to be explicit

spec/unit/hanami/router/url_spec.rb
```diff
@@ -3,13 +3,18 @@
 RSpec.describe Hanami::Router do
   let(:router) do
     e = endpoint
+    # rubocop:disable Layout/ExtraSpacing
```
cllns, line 6: Can we remove this and just comply with the rule instead? It's not as clean looking but it's still fine, especially since this is just a spec

lib/hanami/router.rb
```diff
@@ -878,6 +882,12 @@ def prefixed_name(name)
       @name_prefix.relative_join(name, PREFIXED_NAME_SEPARATOR).to_sym
     end
 
+    # @since 2.3.0
+    # @api private
+    def underscored_name(name)
```
cllns, line 887: It doesn't look like prefixed_name is used anywhere else, and it's defined as a private method + private API, so can we just push the underscoring into that method? It'll save some coercion back-and-forth between symbol and string, but more importantly, it keeps thing simple and doesn't overly abstract helper methods.

cllns (changes requested):
Nice! Thanks for doing this. A few small comments but this looks great :)

lib/hanami/router.rb
svoop: Well spottet, @cllns !

Actually, I quote `-` inside square brackets always and on purpose: Not everybody is experienced enough with Regexp to know and if you add another acceptable character to the front of the list, the unquoted `-` silently flips from being a literal to a range operator. And if this someone is out of luck, the resulting range includes the literal `-` and therefore the tests might not even fail. This quite likely wouldn't be a huge problem here, but it could introduce a difficult to spot bug or even open a security gap elsewhere. Thus, better safe than sorry.

But feel free to insist or change it if you prefer dropping the quote.

spec/unit/hanami/router/url_spec.rb
svoop: Sure thing, removed it.

lib/hanami/router.rb
svoop: Absolutely, it now reads:

```ruby
    def prefixed_name(name)
      @name_prefix
        .relative_join(name, PREFIXED_NAME_SEPARATOR)
        .gsub(NAME_UNDERSCORE_REGEXP, "_")
        .to_sym
    end
```

Hope, the multi-line notation with leading dots  is okay, opinions and tastes differ on this one.

Rubocop appears to be okay with it, but it barks at the constants being declared following `private`:

> Useless private access modifier for constant scope.

How do you deal with such situations?

* Move `private` or exclude the cop rule?
* Mop this up as part of a PR like this one or do a fresh PR for it?

cllns (changes requested):
Thanks for the changes! Looking at this again, I notice we use `prefixed_underscored_name` only once, and pass it directly to `add_named_route`. What if we just do the normalizing in there? Maybe calling is `normalized_as` or even just shadow `as` with the prefixed and underscored version? WDYT?

svoop:
@cllns You're right, the method is only used once in `add_route`, but the resulting `as` is also used a few lines down for the inspector, so moving the normalization to `add_named_route` won't cut it. The normalization could, however, be moved up into `add_route` which adds 3 LOC to a total of 23 LOC. [Looks like this in the end](https://github.com/svoop/hanami-router/blob/e0120c62d5dc15cd6fdacabdf776f0787f5906f8/lib/hanami/router.rb#L794), Rubocop is okay with it.

I'm not sure I understand how you'd "shadow `as`" since it's a local variable and my understanding of shadowing is more along the lines of instavar getters. Could you elaborate this approach?

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 Rack compatible HTTP router for Ruby
 
+## [Unreleased]
+
+### Changed
+
+- [Sven Schwyn] Convert dash to underscore in URL helper names (#280)
```
cllns, line 9:
```suggestion
- [Sven Schwyn] Convert -, +, ~, and . to underscore in URL helper names (#280)
```

svoop: Done... here and equally in the commit message.

cllns (approved):

lib/hanami/router.rb
```diff
@@ -874,8 +878,11 @@ def prefixed_path(path)
 
     # @since 2.0.0
     # @api private
```
timriley, line 880: Let's reset this to `@since x.x.x` since we've changed the method name.

TBH, I'm in favour of dropping the `@since` on anything that's marked `@api private`. It just adds unnecessary busywork for us. This is a change for another time, though :)

timriley (approved):
Nice work, @svoop! One small request for change, but am happy to see you merge it after that 😄

lib/hanami/router.rb
timriley: Oh, I also notice that this is actually `private` in terms of visibility too. Yeah, I'd definitely support all the `@since` and `@api` tags being removed from these methods.

lib/hanami/router.rb
timriley: I really don't like that rubocop rule. If we haven't disabled it yet, we should.

lib/hanami/router.rb
svoop: I've reset it as per your suggestion and created an issue on the mother ship repo for dropping all `@since` for private APIs at some point: https://github.com/hanami/hanami/issues/1516

PR #285, opened 2025-08-26
Add Rack 2/3 to CI build matrix

cllns:
The Rack 2 tests are all failing (proving this is a necessary change!), due to Content-Length header being present when we expect it to be empty. That change was made in this commit: https://github.com/hanami/router/pull/277/commits/2323a3669b0a1053cda1840143518c370cb8777f, and related to this comment: https://github.com/hanami/router/pull/277#issuecomment-3142177837

~I think we don't want to change what we were doing for Rack 2, so we should set the response to have a Content-Length for Rack 2, and not have one for Rack 3? WDYT @timriley~

Update: so the issue was only for block responses. Rack 2 does set Content-Length for blocks, while [Rack 3 does not](https://github.com/rack/rack/blob/b68251c03788ff39d4a4b25424df7360426e4afd/lib/rack/response.rb#L114-L117), so I added a helper method for that. 

Note that there's an optimization that could be made in the implementation, and just avoid rack_headers altogether since they have the same conditional, with different behavior, so it could be simplified as:

```ruby
if Hanami::Router.rack_3?
  Rack::Headers.new
else
  {"Content-Length" => body.length.to_s}
end
```

But I think keeping it consistent with the others in the file, i.e. always using the `rack_headers` helper, is simpler to follow, though it's unoptimized. Since it's just test code (and negligible), I think we should keep it as-is.

Can you confirm this is correct @kyleplump @timriley ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ jobs:
           - "3.3"
           - "3.2"
           - "3.1"
+        rack:
+          - "~> 2.0"
+          - "~> 3.0"
     steps:
       - uses: actions/checkout@v4
       - name: Install package dependencies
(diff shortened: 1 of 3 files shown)


timriley (approved):
Thanks @cllns, this looks right to me! Really appreciate the help getting the matrix into place too. I hope to get https://github.com/hanami/controller/pull/477 merged tomorrow too and we can get it in place for hanami-controller too.

PR #286, opened 2025-10-29
Improve route naming flexibility

timriley:
Make route naming more flexible.

- Provide `as:` to `scope` to customise its route name prefix. This previously used the URL portion of the scope as the prefix in all cases. This is not good for complex scopes, e.g. `"cafes/:cafe_id"`.
- When naming routes with `as:`, provide a `["prefix", "suffix"]` array to allow the prefix to be placed at the very beginning of the generated route name. This is a new capability, and it's relevant when using routes within scopes. For those routes, their names are always prefixed by the names of their scopes, which means it's never been possible to put some portion of the route name at the very beginning of the string. It's important to be able to do this for creating route names with a leading verb, like "new_deeply_nested_thing". To do this, you can now create a route with `as: [:new, :thing]`. See the tests in this PR for examples.

I chose the `["prefix", "suffix"]` array syntax because it is concise and it still reads quite well when given a pair of names.

Aside from this being useful in its own right, this also supports the `resources` routing helpers being added in https://github.com/hanami/hanami/pull/1542.

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -929,9 +929,9 @@ module Hanami
       name = Array(name)
 
       if name.size < 2
-        [nil, name.first.to_s]
+        [nil, name.first&.to_s]
       else
-        [name.first.to_s, name[1..].join("_")]
+        [name.first&.to_s, name[1..].join("_")]
       end
     end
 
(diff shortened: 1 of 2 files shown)


parndt:
@timriley nice - I think it's worth adding another rdoc example for this which will describe what will be produced, wdyt?

timriley:
@parndt Good idea. Done!

parndt (approved):

PR #287, opened 2025-11-03
Rewind the input when body parsing

timriley:
Otherwise, attempting to read the body for non-parseable requests (i.e. anything besides JSON and multipart) would leave the body already-read, and therefore empty for downstream consumers, like our own core router code, which will later try to read standard form-encoded params from the body.

This is a refinement of the Rack 2/3 compatibility work that we did in #277.

### DIFF
--- a/lib/hanami/middleware/body_parser.rb
+++ b/lib/hanami/middleware/body_parser.rb
@@ -39,6 +39,8 @@ module Hanami
       end
 
       def call(env)
+        return @app.call(env) if env.key?(Router::ROUTER_PARSED_BODY)
+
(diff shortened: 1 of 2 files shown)


timriley:
> Since Rack 3 adoption has been so slow, do you foresee supporting both for a long time?

@alassek I reckon we maintain support both versions for a year, or at least two additional Hanami releases (i.e. 2.4 and 2.5), whichever is longer.

Rack 3 has been out for a while now, we're pretty late in adopting it. A year feels like a pretty generous period for people to figure out their transition.

PR #301, opened 2026-04-16
Refactor `recognize` to take params as kwarg

### LINKED ISSUE #271: Router.recognize sets GET to the verb of POST routes
I have a router something like:
```ruby
get '/images', to: 'images.index'
post '/images', to: 'images.upload'

get '/images/:id', to: 'images.show'
post '/images/bulk', to: 'images.upload'
```

[My tool analyses Hanami apps](https://github.com/exoego/rspec-openapi) using `router.recognize`.
I noticed that `recognize` method returns a `RecognizedRoute` whose `verb` is always `GET` even for `POST` or other methods.

```ruby
route = router.recognize("/images/bulk", method: "POST")
expect(route.routable?).to be(true)
expect(route.verb).to eq("POST") # => GET returned 😱 
```

## Repro

Clone https://github.com/hanami/router/pull/270
And run `bundle exec rspec`.

cllns:
Fixes #271. 

`Router#recognize` was silently ignoring the `method:` kwarg when called without an explicit `params` hash, returning the verb as GET for POST routes and often picking the wrong route.

Root cause: the old signature `recognize(env, params = {}, options = {})` absorbed `method: :post` into the positional `params` argument, leaving `options` empty when it was handed to `Rack::MockRequest.env_for`.

Fixes the bug by moving to an explicit kwargs signature:

```ruby
# Before
def recognize(env, params = {}, options = {})

# After
def recognize(target, params: {}, **options)
```

I also renamed `env` to `target` to reflect that the argument accepts a path, a route name, or a Rack env hash.

This is a breaking change, so will need to be for hanami-router v3. It seems like a good time to fix this "the right way" instead of making a backward-compatible fix that we'd have to continue to maintain. This new approach is much more explicit.

There's a slight behavior change in that we double splat the keyword args, so they're converted to symbols, so e.g. "HTTP_AUTHORIZATION" will get converted to a symbol, which Rack drops. There's a very simple fix though: callers who need this can build the env explicitly via Rack::MockRequest.env_for and pass it in as the target.

_I worked with Claude Code to get this fixed, refactored, and all of the documentation updated._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,34 @@ Rack compatible HTTP router for Ruby.
 
 ### Changed
 
+- **BREAKING**: `Router#recognize` signature changed from `(env, params = {}, options = {})` to `(target, params: {}, **options)`. Rack env options (like `method:`) are now kwargs, and named-route variables must be passed under the `params:` keyword.
+
+    Migration:
+
+    ```ruby
+    # Before
+    router.recognize(:book, id: 23)
+    router.recognize("/books/23", {}, method: :post)
+
+    # After
+    router.recognize(:book, params: {id: 23})
+    router.recognize("/books/23", method: :post)
+    ```
+
+    String-keyed Rack env entries (e.g. `"HTTP_AUTHORIZATION"`) can no longer be passed through `recognize` directly, since kwargs only accept symbol keys. Build the env explicitly and pass it as the first argument instead:
+
+    ```ruby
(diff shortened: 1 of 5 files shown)


katafrakt:
Since this is a breaking change, I feel like we kind of missed the window in 3.0 release. However, at the same time reading the original issue I think this is not a bug at all (?). The `:method` option was not silently ignored, it was `router.recognize` that was called incorrectly.

Perhaps we should document this nuance better, but I don't think it justifies a breaking change like this. WDYT?

PR #302, opened 2026-04-17
Require explicit `code:` on redirect, add helper methods

### LINKED ISSUE #283: Reconsider default redirect code
Our default redirect is [301: Moved Permanently](https://en.wikipedia.org/wiki/HTTP_301)

https://github.com/hanami/router/blob/50826c98fa9c98ab9a1bfd1de7c99e35370b42e7/lib/hanami/router.rb#L710

In browsers, this is often cached and it can be hard to break that cache. I think we should consider switching our default to [302: Found](https://en.wikipedia.org/wiki/HTTP_302), which is a temporary redirect.

This is a breaking change so it'd have to be for a 3.0 release. The other option would be to keep 301 as the unconfigured default, then make this a configurable value, then generate new apps with the configuration set. But that seems like a lot of work for something so minor.

cllns:
Closes #283

Removes the default 301 from redirect and requires `code:` explicitly. 

Adds `permanent_redirect` (301) and `temporary_redirect` (302) helpers for the common cases. 

This is a breaking change, since it requires `code:` kwarg (so will have to target v3.0+), whereas now we default to 301. I'm also open to having 302 as the default and *not* requiring it, but «there are worse things than being explicit». 

_I worked with Claude Code to implement this and update the docs accordingly._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ Rack compatible HTTP router for Ruby.
 
 ### Added
 
+- `#permanent_redirect` and `#temporary_redirect` helpers for 301 and 302 responses. (@cllns in #283)
+
 ### Changed
+
+- **BREAKING:** `#redirect` now requires an explicit `code:` keyword argument. Use `#permanent_redirect` or `#temporary_redirect` for the common cases, and pass `code:` to `#redirect` for less common codes (303, 307, 308). (@cllns in #283)
 
 ### Deprecated
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Hanami::Router.new do
   get "/dashboard",   to: Dashboard::Index
   get "/rack-app",    to: RackApp.new
 
-  redirect "/legacy", to: "/"
+  permanent_redirect "/legacy", to: "/"
 
   mount Api::App, at: "/api"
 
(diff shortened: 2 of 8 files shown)


kyleplump (approved):
LGTM @cllns! :+1:

timriley:
Thought experiment: what about calling these `redirect_permanent` and `redirect_temporary`? Putting the "redirect" first in the name might make things clearer when scanning routes?

e.g.

```ruby
Hanami::Router.new do
  get "/redirect_destination", to: ->(env) { [200, {}, ["Redirect destination!"]] }
  redirect_permanent "/legacy",     to: "/redirect_destination"
  redirect_temporary "/today",      to: "/redirect_destination"
end
```

To me, this feels nicer than:

```ruby
  permanent_redirect "/legacy",     to: "/redirect_destination"
  temporary_redirect "/today",      to: "/redirect_destination"
```

It's the redirect itself that's the operative thing, the _nature_ of the redirect feels secondary, which is why I think inverting the name feels more "right". They would also fit better with the naked use of `redirect` (which now requires the trailing `code:`).

What do you think?

cllns:
Yea that makes sense to me! Just did that in [4eb4511](https://github.com/hanami/hanami-router/pull/302/commits/4eb4511e5d233240e0047ebbfb4955aab0d5bd34)

timriley (approved):
Thanks @cllns!

PR #304, opened 2026-04-27
Add array variables extraction + injection around Mustermann expand method

### LINKED ISSUE #303: Error while trying to generate an URL with an array query parameter
In Hanami 1 I used to generate URLs sometimes with arrays, like that : `routes.path(:my_route, :foo => ["bar", "baz"])`
It generated urls with query string like this : `/my/route?foo[]=bar&foo[]=baz` which are parsed as array by Rack.

In Hanami 2, when I try to generate the same kind of path, Mustermann chokes on it: 
```ruby
undefined method 'gsub' for an instance of Array (NoMethodError)
11:54:09 app.1 |   .rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/gems/uri-1.1.1/lib/uri/rfc2396_parser.rb:291:in 'URI::RFC2396_Parser#escape'
11:54:09 app.1 |  .rbenv/versions/3.4.5/lib/ruby/gems/3.4.0/gems/mustermann-3.0.4/lib/mustermann/ast/translator.rb:126:in 'Mustermann::AST::Translator#escape'
```

On the other hand, Rack understands it correctly as an array when you build the URL by hand, `request.params[:foo]` contains `["bar", "baz"]`

## To Reproduce

Call `routes.path(:my_route, :foo => ["bar", "baz"])`

## Expected behavior

Should generate the path `/my/route?foo[]=bar&foo[]=baz`, but raise an exception instead, as a String was expected

inouire:
Trying to fix #303 

It's a bit tricky to handle array query variables as Mustermann handled expansion by itself.

We have to extract array variables of its scope otherwise it will choke on non-array variables

We can add serialized arrays afterwards so that Mustermann is not aware of them

### DIFF
--- a/lib/hanami/router/url_helpers.rb
+++ b/lib/hanami/router/url_helpers.rb
@@ -28,9 +28,16 @@ module Hanami
       # @since 2.0.0
       # @api private
       def path(name, variables = {})
-        @named.fetch(name.to_sym) do
+        safe_variables = variables.reject do |key, value|
+          value.is_a?(Array)
+        end
+        array_variables = extract_and_patch_array_variables(variables)
+
+        expanded_path = @named.fetch(name.to_sym) do
           raise MissingRouteError.new(name)
-        end.expand(:append, variables)
+        end.expand(:append, safe_variables)
+
+        append_array_variables(expanded_path, array_variables)
       rescue Mustermann::ExpandError => exception
         raise InvalidRouteExpansionError.new(name, exception.message)
       end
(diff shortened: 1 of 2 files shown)


inouire:
Or maybe would it be Mustermann responsibility to do it in the expand function?

inouire:
@timriley do you think it can be part of next release? Thanks

timriley:
@inouire thank you for the nudge! Yes, this will absolutely be going in. I'll make sure I review it soon. We'll have a 3.0 rc coming out in a few weeks, and this will be included.

timriley:
Thanks @inouire, this is great! I've done a little refactoring (see the commit pushed above) and now I'll merge this in. We'll ship this with 3.0. A nice improvement!

inouire:
Amazing, thank you. Glad it helped!

PR #240, opened 2022-11-17
Don't parse request body when Body Parser already parsed it

jodosha:
Fixes https://github.com/hanami/router/issues/237

PR #305, opened 2026-05-05
Fix nested form body parsing happening for all content types

cllns:
I have a JSON API Hanami app and when benchmarking it, I had some generated code that POSTed with "%" in a field, as in "Improves speeds by 50%". This surprisingly raised a `Rack::QueryParser::InvalidParameterError`.

This is because we run `Rack::Utils.parse_nested_query` on the request body for _every_ request, regardless of Content-Type. This causes JSON bodies to leak into router.params (as evidenced by the JSON regression test failure added here: `{from: "ui", "{\"foo\":\"bar\"}": nil}`), while bodies containing a literal `%` (#237) or multipart binary bytes (#296, other regression test) raise `Rack::QueryParser::InvalidParameterError` because it's being parsed as form input (nested query) and the % is interpreted as malformed URL encoding.

#237 was closed with #240 as the workaround, but that was at a different time in the frameworks history: when we required JSON parsing to be done via a body parser in this repo. Now we're more flexible and parse it later on, so the problem gets triggered upstream. This fixes it at its proper place, only using `::Rack::Utils.parse_nested_query` for forms.

Closes #296

### DIFF
--- a/lib/hanami/router.rb
+++ b/lib/hanami/router.rb
@@ -1040,16 +1040,24 @@ module Hanami
       params ||= {}
       env[PARAMS] ||= {}
 
-      if !env.key?(ROUTER_PARSED_BODY) && _form_urlencoded?(env) && (input = env[::Rack::RACK_INPUT]) and input.rewind
-        env[PARAMS].merge!(::Rack::Utils.parse_nested_query(input.read))
-        input.rewind
-      end
-
+      _merge_form_urlencoded_body!(env)
       env[PARAMS].merge!(::Rack::Utils.parse_nested_query(env[::Rack::QUERY_STRING]))
       env[PARAMS].merge!(params)
       env[PARAMS] = Params.deep_symbolize(env[PARAMS])
 
       env
+    end
+
+    # @since x.x.x
+    # @api private
+    def _merge_form_urlencoded_body!(env)
+      return if env.key?(ROUTER_PARSED_BODY)
+      return unless _form_urlencoded?(env)
+      return unless (input = env[::Rack::RACK_INPUT])
+
+      input.rewind
+      env[PARAMS].merge!(::Rack::Utils.parse_nested_query(input.read))
+      input.rewind # leave the stream readable for downstream consumers
     end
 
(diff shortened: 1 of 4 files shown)


lib/hanami/router.rb
```diff
@@ -1040,18 +1040,36 @@ def _params(env, params)
...
+      input.rewind # leave the stream readable for downstream consumers
+    end
+
+    # @since x.x.x
+    # @api private
+    def _form_urlencoded?(env)
+      content_type = env[CONTENT_TYPE]
+      return false unless content_type
+
+      content_type == FORM_URLENCODED_MEDIA_TYPE ||
```
timriley, line 1078:
```suggestion
      content_type.downcase! == FORM_URLENCODED_MEDIA_TYPE ||
```

Feels like we should do this for maximum compatibility, and preservation of our existing behaviour. `Rack::Request#media_type` returns a downcased string courtesy of its internal `MediaType.type`. And if we're aiming to follow [the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type) strictly, it does say that these values are case insensitive.

lib/hanami/router.rb
```diff
@@ -1040,18 +1040,36 @@ def _params(env, params)
...
+        content_type.start_with?("#{FORM_URLENCODED_MEDIA_TYPE};")
```
timriley:
```suggestion
      content_type.downcase!.start_with?(FORM_URLENCODED_MEDIA_TYPE)
```

In fact, I wonder whether we could actually just simplify it to this?

It'd work fine for both `type/subtype` as well as `type/subtype;parameter=value`.

timriley (approved):
Nice find and nice fix, @cllns! Kudos on the perf improvements along the way too. :)

I've left a couple of small suggestions which you might want to consider before merging. I'm happy for you to merge whenever you're ready :)

timriley:
Made a few tidy-ups and merged this one. Thanks for finding and fixing it, @cllns!

PR #306, opened 2026-08-16
Add a way to grab the routes on a given router for reading back

afomera:
As part of my work to improve the Hanami error pages, the 404 error page would be nice to include routes so you could see what existed etc. But there is no way to return given routes for a router without jumping through hoops and creating inspectors and whatnot. So this gives a public API that exposes the details for us but also perhaps a hanami LSP down the road. 👀


katafrakt:
To me it feels like we are duplicating 80% of the role of inspector here. I don't think it's good. I'm not sure why Luca was so conservative about saving the routes in an instance variable, but in a Chesterton's fence spirit, I personally wouldn't touch it.

Reading through the problem described, I was able to get the same results with the following code snippet, without any changes to the framework:

```ruby
require "hanami/router/inspector"
class InternalInspector < Hanami::Router::Inspector
  def routes = @routes.each
end

Hanami.app.router(inspector: InternalInspector.new).inspector.routes.each do |route|
  puts "#{route.http_method} #{route.path}"
end
```

Perhaps that's few enough hoops?

katafrakt:
Oh, I think I found the answer in #212:

> Ensure that the routes inspection is lazy, so the memory won't grow up unnecessary at the startup time when inspection isn't triggered.

afomera:
@katafrakt great callout, I had hoped to do similar there, but the main thing there was Slice.router memorizes with the inspector and it only is honored on the first call, but in the webconsole case it wouldn't be set then at the time of rendering the errors.

There might be a compromise we could do where it keeps the #212 laziness intact and not touch Hanami::Router and add an attr_reader to Inspector and have Hanami's load_router pass the inspector only when render_detailed_errors? is true. Which is probably the more optimal solution. 

Thoughts on that approach @katafrakt? 

Edit: one of things is in a fresh console the inspector loads since the memoization runs with the inspector as passed, but in the request situation it wouldn't be there because its already built without that inspector

katafrakt:
I though `Hanami.app.router(inspector: InternalInspector.new)` will create a new instance of the router, with swapped inspector, but I don't know why I thought that. In this case, I think we could attach Hanami-specific inspector in `Slice::Router`, under the conditions you mentions. Although I would be keen to hear Tim's opinion about this.

PR #307, opened 2026-08-19
Add conditional support for QUERY HTTP method

katafrakt:
Defined in [RFC 10008](https://datatracker.ietf.org/doc/html/rfc10008), QUERY is a new proposed HTTP method. Head Rack and Puma already support it. This adds support to Hanami Router conditionally, based on Rack's support.

Screenshot_20260819_214643


.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         rack:
           - "~> 2.0"
           - "~> 3.0"
+          - "head"
```
katafrakt, line 34: this would need to be supported by repo sync, I believe, but not sure if we want to merge it before there's a Rack proper release with the support (in which case the `"~> 3.0"` about should be enough and provide support for QUERY).

katafrakt: nevermid, it already is

lib/hanami/router.rb
```diff
@@ -224,6 +224,22 @@ def get(path, to: nil, as: nil, **constraints, &blk)
...
+      # Defines a route that accepts QUERY requests for the given path.
+      #
+      # @param path [String] the relative URL to be matched
+      # @param to [#call] the Rack endpoint
+      # @param as [Symbol, Array<Symbol>] a unique name for the route, or a ["prefix", "name"] array,
+      #   to add a prefix to the name when nested within scopes.
+      # @param constraints [Hash] a set of constraints for path variables
+      # @param blk [Proc] the anonymous proc to be used as endpoint for the route
+      #
+      # @since 0.1.0
```
timriley: Is `@since 0.1.0` what we actually want here? If this is a legitimately new method, maybe `@since 3.1.0` is what we want?

timriley (approved):
Looks fantastic, thanks @katafrakt!

I left one question about the API docs. Happy for you to merge once you've resolved that :)

lib/hanami/router.rb
katafrakt: Certified human copy-paste error. No LLM would do such a silly mistake. Fixed, thanks for spotting. Should we merge it already or wait for Puma support?

timriley: @katafrakt Correct me if I'm wrong, but I think there's basically no disruption if we merge now? And it means as soon as someone upgrades (eventually) to a compatible Rack+Puma, suddenly everything works? If so, then I think we can merge now :)

katafrakt: That's correct. Merging then.

PR #309, opened 2026-09-12
Remove unused private `with` method

katafrakt:
This was used by `to_inspect` method, which was then removed in 1f7d40ab14f2c48525b5b9c4895bde64628b7bb9.

I thought that maybe it is used by slice router in Hanami, but not only it's not, but also this method does not work, as slice router makes additional options required when building a new object.

An alternative route (pun unintended) could be to make this method public. Then we'd have to use a trick similar https://github.com/dry-rb/dry-cli/pull/167, I believe, so subclasses can redefine the list of required options. This could then help @afomera in #306 where the problem is inability to swap the inspector, because router is already frozen. With public `with` this could be achieved by creating a new object using `router.with(inspector: new_inspector)`. But not sure it's worth it - if yes, I can repurpose this PR for it.


timriley (approved):
Thanks for noticing this, @katafrakt! I reckon we remove the unused code now, and if there's reason to introduce something like this (for e.g. #306), then we can just do whatever is most helpful for that. We can always see the old code via the commit history anyway :)

katafrakt:
Sure, I just kind of wanted to raise awareness that such idea apparently existed before and maybe could help, but no need to keep this function if we don't know whether its useful or not.

timriley:
I'm glad you shared it, @katafrakt! ❤️

